PR #6, opened 2018-10-31
Fixes broadcast_as attribute error

### LINKED ISSUE #5: numpy AttributeError in hcipy/field/coordinates.py
I get 'numpy' has no attribute 'broadcast_as' when running the adaptive_optics example. It originates in hcipy/field/coordinates.py line 176.

dskleingeld:
This should resolve #5, it also fixes some non-tuple indexing warnings that fire when running the adaptive_optics example.


ehpor:
Thank you. The first was already noted and fixed in a separate branch: d2d8cd3. The second one was still on my todo list.

PR #15, opened 2019-05-03
Create a Segmented Mirror simulator

ivalaginja:
I wrote a `SegmentedMirror()` class that uses one of the segmented aperture generation functions to create a segment simulator. The class is based on the `DeformableMirror()` class and the segment behavior was mostly adopted from Poppy's `HexSegmentedDeformableMirror()`, but adapted to the HCIPy propagation and imaging framework.

In principle, the segmented mirror should be able to take any segmented aperture from HCIPy and make a segment simulator out of it. For this first development case though, I had to copy one of the aperture functions and make it return its segment positions, as they are the basis for the PTT modes on the segments. This will need to be generalized to make it work on arbitrary segmented apertures. The modified aperture function is included in the `segmented_mirror` module.

I also pushed a notebook that demonstrates and tests the `SegmentedMirror()` class and compares it to the Poppy equivalent. I wasn't sure where to put it so I created a new directory `notebooks`. Please advise on how to handle this @ehpor. I pushed the notebook with cleared outputs in order to not clutter the repo with large files, so you'll have to run it locally to see the figures.

And lastly, there are definitely some minor details that need to be changed, like the parameter inputs for grid size, telescope size etc.

@ehpor I created this Draft pull request because it might make communication easier and we can add inline comments to the code itself if needed. I already added comments where I feel like improvements and changes need to be done.


hcipy/optics/segmented_mirror.py
```diff
@@ -0,0 +1,237 @@
...
+
+from .optical_element import OpticalElement
+from ..aperture import circular_aperture, hexagonal_aperture, make_segmented_aperture
+from ..field import Field, make_hexagonal_grid, make_pupil_grid, evaluate_supersampled
+from ..io import write_fits
+from ..plotting import imshow_field
+
+
+pupil_size = 1024
+PUP_DIAMETER = 15.   # m
```
ivalaginja: These need to go and either be an input parameter or inside the function/class.

hcipy/optics/segmented_mirror.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+def get_atlast_aperture(normalized=False, with_segment_gaps=True, segment_transmissions=1, write_to_disk=False, outDir=None):
```
ivalaginja: It would be great to refactor either the segmented aperture functions or the SegmentedMirror() class such that we can use the original aperture functions and don't need to create a duplicate just because we need to return the segment positions.

hcipy/optics/segmented_mirror.py
```diff
@@ -0,0 +1,237 @@
...
+        return hexagon(grid.rotated(np.pi/2))
+
+    segmented_aperture = make_segmented_aperture(segment, segment_positions, segment_transmissions)
+
+    def func(grid):
+        res = segmented_aperture(grid)
+
+        return Field(res, grid)
+
+    # Save pupil to disk, as pdf and fits
```
ivalaginja: The saving to disk can probably be taken out, I just realized I left it in there because I needed it for my code.

hcipy/optics/segmented_mirror.py
```diff
@@ -0,0 +1,237 @@
...
+
+        x, y = self.input_grid.coords
+
+        self._seg_mask = np.zeros_like(x)
+        self._seg_x = np.zeros_like(x)
+        self._seg_y = np.zeros_like(y)
+        self._seg_indices = dict()
+
+        pupil_grid = make_pupil_grid(dims=npix, diameter=PUP_DIAMETER)
+        aper_num, seg_positions = get_atlast_aperture(normalized=False,
```
ivalaginja: It doesn't seem like a good solution to evaluate the aperture inside the class, even if it works for now, because 1) I already pass in the aperture generator and the segment positions when I instantiate the class and 2) it makes the segmented mirror only work for one specific aperture, which is very restrictive. This definitely needs to be changed.

hcipy/optics/segmented_mirror.py
```diff
@@ -0,0 +1,237 @@
...
+        for i in self.segmentlist:
+            wseg = np.where(self._seg_mask == i)
+            self._seg_indices[i] = wseg
+
+            cenx, ceny = self.seg_pos.points[i - 1]
+
+            self._seg_x[wseg] = x[wseg] - cenx
+            self._seg_y[wseg] = y[wseg] - ceny
+
+            # Set gaps to zero
```
ivalaginja: When I initially tested this, the gaps between the segments had super high values that would mess with the actual phase ramps on the segments. I solved it by defining a threshold and setting anything above it to zero, but I feel like there might be a more robust solution to this.

hcipy/optics/segmented_mirror.py
ehpor: Yes, this should be an optional parameter `return_segment_positions` for every segmented aperture, starting with `make_segmented_aperture()`.

hcipy/optics/segmented_mirror.py
ehpor: Yes, take this out.

hcipy/optics/segmented_mirror.py
ehpor: You can pass in the `aper_num` and `segment_positions` as arguments to the class. You can then also get the pupil grid from that `aper_num`. Also rename `aper_num` to `subaperture_indices`.

hcipy/optics/segmented_mirror.py
ehpor: Agreed.

hcipy/optics/segmented_mirror.py
ehpor: The problem here is that for each pixel it may not be perfectly clear to which segment it belongs. This should probably be changed in the `make_segmented_aperture()` function. Maybe this function should be able to return the `subaperture_indices`?

ehpor:
More general comments:
* The notebook is pretty good already. We should refactor this to serve as a tutorial. The documentation consists of three parts: a getting started guide (least amount of knowledge to start working with HCIPy), a tutorial section (showcasing pretty examples) and a full API documentation (documentation for every function/class). This notebook seems perfect for a tutorial.
* Should it be possible to include more aberrations than just tip-tilt and piston? For example defocus?
* Can this class derive from DeformableMirror and precalculate the tip-tilt-piston modes for each segment. This would cost a lot of memory, so I love the approach here. (Although the same thing can be said for the sparse modes for a DM influence function set. Maybe a small refactoring of `ModeBasis` is in order here, to allow sparse mode bases.
* If on-the-fly mode calculation is needed. Maybe some caching of the surface would be better a good thing to have, especially for broadband simulations.

ivalaginja:
@ehpor Did you have a chance yet to commit your changes and updates?

ehpor:
I didn't test everything yet, so there may be a few typos.

I added caching of the surface and tried speeding up the atlast aperture function. I also removed subsampling of the segment indices which was probably the origin of the gaps between segments having high values which required filtering. I didn't remove the filtering, but it should work without it, if I'm not mistaken.

hcipy/optics/segmented_mirror.py
ivalaginja: Should it really be `make_segmented_aperture()` that has a flag for returning the segment positions? The segment positions are an input to that function, so wherever it is used, they will already be available - would there still be a reason to have `make_segmented_aperture()` return the segment positions nevertheless?

ehpor: Agreed. I realized this as well, while letting make_atlast_aperture() return the segment positions. Ignore the last part of that sentence.

ivalaginja: OK. I added the `return_segment_positions` keyword for the HiCAT and LUVOIR aperture functions, they will come up in the next push.

hcipy/optics/segmented_mirror.py
ivalaginja: This in-situ aperture evaluation is here because the segment transmissions used here provides the segment identification for the mirror. I fell like that could also be something returned by the aperture directly, but I’ve realized we’ve started talking about adding a lot of function returns for the segmented apertures to accommodate the `SegmentedMirror()` functionalities. Should we just proceed with this? Can you think of a more elegant solution for passing in the segment numbering? Would it make sense of adding a class that uses `Field` as its base class which we can use to pass apertures around that are already evaluated on a grid, and that have attributes like its segment positions, diameter, segment numbering etc.?

hcipy/optics/segmented_mirror.py
ivalaginja: `PUP_DIAMETER` is currently used in a couple of instances in `SegmentedMirror()`, specifically in the method `_setup_grids()`. It scales the segment gaps and creates a pupil grid for the aperture with the *numbered* segments. I'm unsure of how to get rid of this global variable as long as we need that scaling. See also my comment about the in-situ evaluation of the aperture.

ivalaginja:
@ehpor everything (at least in the notebook) still works fine after your commit. My commit only has minor stuff: I added a return argument for segment positions in the segmented apertures in HCIPy, so `get_atlast_aperture()` can be deleted as soon as we solve the problem of the in-situ aperture evaluation in the `SegmentedMirror()` class, since it needs to be agnostic to different segmented apertures.

hcipy/optics/segmented_mirror.py
```diff
@@ -0,0 +1,227 @@
...
+		segment_gap /= pupil_diameter
+		pupil_diameter = 1.0
+
+	segment_positions = make_hexagonal_grid(segment_circum_diameter / 2 * np.sqrt(3), num_rings)
+	segment_positions = segment_positions.subset(lambda grid: ~(circular_aperture(segment_circum_diameter)(grid) > 0))
+
+	hexagon = hexagonal_aperture(segment_circum_diameter - segment_gap, np.pi / 2)
+
+	def segment(grid):
+		return hexagon(grid)
```
ivalaginja: The same thing can be done for the HiCAT and LUVOIR aperture. Should we include that in this PR?

hcipy/optics/segmented_mirror.py
ivalaginja: Alternatively, we could pass the aperture generator and a pupil grid into the `SegmentedMirror()` class, instead of the sampled aperture. This would still lead to the evaluation of the aperture in the class, but it would make it agnostic to other segmented apertures and with caching, the evaluation would happen only once or twice for an instance of the class.

hcipy/optics/segmented_mirror.py
ivalaginja: I added these changes to the HiCAT and LUVOIR apertures in commit 7dde6f31ef55bb5d4bc40c415d8a5813bd804814.

ivalaginja:
I changed the code such that there is no need for the custom aperture function anymore, since there is now the option of returning the segment centers from the segmented aperture functions themselves. There is still an aperture evaluation inside the segmented mirror class though, I am currently trying to come up with a way of getting around that.

ivalaginja:
I changed the class such that the input is an indexed aperture instead of the transmission aperture used for the propagation. This eliminates the need to evaluate the aperture inside the class in order to get the segment indices. It increases the overhead to instantiate the `SegmentedMirror()` class, but it makes it agnostic to any sort of segmentation.

I adjusted the notebook for it. Suggestions for improvements appreciated.

hcipy/optics/segmented_mirror.py
ivalaginja: This has been resolved with the refactoring of the class for inputs of indexed apertures.

hcipy/optics/segmented_mirror.py
ivalaginja: See https://github.com/ehpor/hcipy/pull/15#issuecomment-495401660

hcipy/optics/segmented_mirror.py
ivalaginja: See https://github.com/ehpor/hcipy/pull/15#issuecomment-495401660

hcipy/optics/segmented_mirror.py
```diff
@@ -0,0 +1,160 @@
...
+		"""
+		self._surface = None
+		return self._coef
+
+	def show_numbers(self):
+		""" Display the mirror pupil with numbered segments.
+		"""
+		imshow_field(self.ind_aper)
+		for i, par in enumerate(self.seg_pos):
+			plt.annotate(s=i+1, xy=par, xytext=par, color='white', fontweight='bold') #TODO: scale text size by segment size
```
ivalaginja: The text positions themselves need to be scaled by the aperture grid diameter.

This display function will work all right if the segment positions are in the same units like the input indexed aperture. I ran into issues because I would evaluate the indexed aperture on a differently sized grid than it was created with, while the segments positions are tied to the original grid it was created with. Not sure how to limit this properly - also, it is just supposed to be a convenience function.

ehpor:
I tried a few things with a sparse ModeBasis. This mode basis uses a sparse matrix from scipy to store both the piston, tip and tilt from the segments. This mode basis is then used as the influence functions for a standard DeformableMirror. This means that all functionality of a DeformableMirror is integrated and not duplicated. Also, sparse ModeBasis can be used for more than just segmented mirrors. One can imagine having the influence functions of a large deformable mirror to also be sparse. Storage of the influence functions for large deformable mirrors is then also reduced, and the calculation of the surface is then sped up as well.

hcipy/aperture/generic.py
```diff
@@ -249,7 +253,7 @@ def make_obstruction(aperture):
 	'''
 	return lambda grid: 1 - aperture(grid)
 
-def make_segmented_aperture(segment_shape, segment_positions, segment_transmissions=1):
+def make_segmented_aperture(segment_shape, segment_positions, segment_transmissions=1, return_segments=False):
```
ehpor: One likely change is a different way of getting the segments. Right now, this is part of the make_segmented_aperture() function, but this doesn't really work, as that function is supposed to return a field generator only. Now, if `return_segments` is True, the "Field generator" returns both a Field and a ModeBasis, so technically it is not a Field generator anymore. It may be better to have a separate function, or maybe even better, have the function `make_segmented_aperture()` return both a Field generator and a ModeBasis generator, or something.

PR #24, opened 2019-07-08
Propagator changes

syhaffert:
The polychromatic propagator generator was changed from a specialized function to the general make_polychromatic that is used for all other optical elements. Several minor additions and fixes have been added to the propagators that decrease numerical aliasing issues for large distance and wide angle propagation.


ehpor:
@syhaffert Can you add/modify a test to ensure that the large-distance Fresnel and angular spectrum propagators work as expected? Otherwise, this is mergable in my opinion.

PR #20, opened 2019-07-06
LUVOIR-A and HiCAT apertures and Lyot stops

ehpor:
I think this can now be merged, as they are now finished. Some cleanup needs to be done to remove IDE and build files that crept in.

PR #26, opened 2019-07-09
Refactor LUVOIR aperture function for segment ModeBasis

ivalaginja:
PR #15 introduced the return of a segment ModeBasis to the `make_hicat_aperture()` that I also added to the `make_luvoir_a_aperture()` function here, plus some updates to respective docstrings.

This is in pretty high conflict with PR #20 and will have to be merged very carefully, attention @ehpor. I realized to late that the giant red and green block because of tab vs. 4spaces indentation will not make this any easier, although I am not sure how to deal with that except for retyping the sections. Can definitely change that if someone has advice.

Tested with:
```py
import hcipy

pupil_grid = hcipy.make_pupil_grid(dims=500, diameter=15.)

aper, segments = hcipy.make_luvoir_a_aperture(normalized=False, with_spiders=False, return_segments=True)
aper = hcipy.evaluate_supersampled(aper, pupil_grid, 1)
segments = hcipy.evaluate_supersampled(segments, pupil_grid, 1)

plt.figure(figsize=(10, 10))
hcipy.imshow_field(aper)
plt.show()
```

### DIFF
--- a/hcipy/aperture/realistic.py
+++ b/hcipy/aperture/realistic.py
@@ -63,93 +63,102 @@ def make_magellan_aperture(normalized=False, with_spiders=True):
 def make_keck_aperture():
 	pass
 
-def make_luvoir_a_aperture(normalized=False, with_spiders=True, with_segment_gaps=True, segment_transmissions=1, return_segment_positions=False):
-	'''Make the LUVOIR A aperture.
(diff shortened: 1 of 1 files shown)


ivalaginja:
Actually. @ehpor you did exactly that in #20 so this PR here is duplicate code and can be completely ignored. Apologies.

ehpor:
Yeah, this was the old LUVOIR-A pupil, so I didn't bother writing code to do the segments correctly for that one. My aim is to merge #20 right after #24.

PR #37, opened 2019-08-16
Added functionality for an on-axis plane wave

rahulbhadani:


syhaffert:
Fields are not necessarily electric fields. They could be fields of any physical (even non-physical) parameter like temperature or pressure. The creation of a plane wave (an electric field) should be done with the class Wavefront.  Currently a plane wave can be made with the following code:

```
grid = make_pupil_grid(num_pixels, grid_diameter)
plane_wave = Wavefront(grid.ones())
```

There will always be a grid first because all optical elements will need to know on what grid to evaluate itself. Therefore you can already use the code in the snippet to create a plane wave.

rahulbhadani:
Alright, thanks for the concise explanation.

PR #42, opened 2019-09-17
Use pillow for gif conversion

jamienoss:
Also:

 * Remove literal path delimiters
 * Don't remove dir containing individual frames

PR #43, opened 2019-09-19
Fix regression from #42 for single file GIFs

jamienoss:
Regression from #42 

Signed-off-by: James Noss <jnoss@stsci.edu>


jamienoss:
@ehpor Sorry, I missed this.

PR #61, opened 2020-02-18
Fix the inf. atmos. for variable wind speed

syhaffert:
I found a small bug in the infinite atmospheric layer that prevented a variable wind speed. This has now been fixed by the commit. The code has been tested and works.


ehpor:
Thanks. I knew about this one (Maaike, Steven and I also encountered it individually). I never got around to fixing it.

PR #72, opened 2020-05-13
Vortex fix for apertures smaller than the grid size

syhaffert:
The implementation of the vortex coronagraph assumed that the pupil diameter was the same as the grid diameter. The proposed update fixes this to include pupil diameters smaller than the grid.

I have modified the tests of the vortex and the new implementation still passes them.

### DIFF
--- a/hcipy/coronagraphy/vortex.py
+++ b/hcipy/coronagraphy/vortex.py
@@ -37,7 +37,7 @@ class VortexCoronagraph(OpticalElement):
 		grid_fill_fraction = pupil_diameter / grid_size
 
 		num_airy_outer = input_grid.shape[0] / 2 * grid_fill_fraction
-		q_outer = max(2, max(grid_size / pupil_diameter))
+		q_outer = max(2, max(2/grid_fill_fraction))
 
 		focal_grids = []
 		self.focal_masks = []
@@ -158,10 +158,6 @@ class VectorVortexCoronagraph(AgnosticOpticalElement):
 		AgnosticOpticalElement.__init__(self)
 
 	def make_instance(self, instance_data, input_grid, output_grid, wavelength):
-		#q_outer = 2
-		#num_airy_outer = input_grid.shape[0] / 2
-		#pupil_diameter = input_grid.shape * input_grid.delta
-
 		grid_size = input_grid.shape * input_grid.delta 
 		if self.pupil_diameter is None:
 			pupil_diameter = input_grid.shape * input_grid.delta
@@ -169,7 +165,7 @@ class VectorVortexCoronagraph(AgnosticOpticalElement):
 		grid_fill_fraction = pupil_diameter / grid_size
 
 		num_airy_outer = input_grid.shape[0] / 2 * grid_fill_fraction
-		q_outer = max(2, max(grid_size / pupil_diameter))
+		q_outer = max(2, max(2/grid_fill_fraction))
 
 		focal_grids = []
 		jones_matrices = []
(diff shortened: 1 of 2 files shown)


spbos:
We should probably also check if the VectorVortexCoronagraph is affected by this.

syhaffert:
The vector vortex had the same problem. I fixed it at the same. The vector vortex tests also passed

ehpor:
The VectorVortexCoronagraph has the same problem, but is also fixed by this PR.

BTW, this is a very minor issue. Multiple people have used the VortexCoronagraph with (highly) undersized pupils without problems. There are more severe problems with the current implementation.

What is the reason for using 512 pixels across the pupil for the tests? Doesn't it work for 256 pixels?

spbos:
Cool, thanks

syhaffert:
With 256 pixels I do not pass the test, because the smaller pupil then uses 128 pixels. I increased the number of pixels so that the smaller pupil still had sufficient accuracy to pass the test.

I got into trouble with the implementation because I was trying to implement the SCC with the vortex coronagraph. I wanted to measure the throughput of the pinhole as function of distance. For this I needed to make the grid larger and I got wrapping effects.

What are the more severe problems? Do you mean my fixes or other problems that were already present?

ehpor:
What I mean is that you are *not* seeing wrapping, but rather imperfect masking of the different layers. Here is an example showing the comparison between the current master branch and the vortex_fix branch:

Current master branch:
old
Vortex_fix branch:
new
Pupils:
pupils

Test code:
```python
from hcipy import *
import hcipy
import numpy as np

for i, field_diameter in enumerate([1,2,3,4]):
    pupil_grid = make_pupil_grid(512 * field_diameter, 1)
    focal_grid = make_focal_grid(4, 256)

    pupil_diameter = 1 / field_diameter

    vortex = VortexCoronagraph(pupil_grid, 2, pupil_diameter=pupil_diameter)

    pupil = circular_aperture(pupil_diameter)(pupil_grid)
    lyot = Apodizer(circular_aperture(0.99 * pupil_diameter)(pupil_grid))

    wf = Wavefront(pupil)
    img = prop(lyot(vortex(wf))).power
    img_0 = prop(lyot(wf)).power

    plt.subplot(2, 2, i+1)
    imshow_field(np.log10(img / img_0.max()), vmin=-10, vmax=-6)
    #imshow_field(pupil)
plt.suptitle(hcipy.__version__)
plt.savefig('new.png')
plt.show()
```
+small modifications for the master branch.

This test keeps the number of pixels across the pupil constant, so increases the number of pixels across the pupil_grid. Not doing this results in the bad results that look like wrapping, but actually isn't.

Another clue to this is that the CI tests that you modified, are also passed by the current master branch.

syhaffert:
I see that my 'fix' just switches the size of the spatial resolution with num_airy so numerically for the rest of the code there is no difference.

At first I had issues that the Lyot plane was wrapping, so not the focal plane. But at the moment I do not get that with the most up to date master branch. I was switching between the vortex code and the Lyot coronagraph. For the Lyot it is crucial to select the correct q for the grid of the focal plane mask to not get wrapping. I think I confused these two because I was switching so often between them.

ehpor:
So I can close this, without merging?

FYI: for the Lyot coronagraph, you typically need at least >10 pixels per lam/D, so you should never see wrapping from that unless you are oversizing the pupil_grid by an insane amount.

syhaffert:
Yes this can be closed. Thanks guys!

Issue #75, opened 2020-07-27
NoisyDetector enhancement: subsample field 


spbos:
Currently, when trying to integrate supersampled Wavefronts with the Noisy Detector, you first have to manually subsample the electric field to the correct grid size of the detector. I find this a bit cumbersome and unnecessary. 

I think it would be more intuitive (and convenient) if we add the subsampling also to the integrate function of the Noisy Detector. The code would check if the wavefront grid shape matches the grid shape of the detector. Then, if necessary, it would subsample it to the correct shape. 

@ehpor If you like this idea, I can code it up and create a new pull request this week.

ehpor:
Agreed. But I wouldn't make it adaptive: that may be unpredictable when you change sampling. Put in the amount of oversampling in the `__init__()`, and let the `integrate()` function do the subsampling always by this factor. That would be cleaner, I think. Go for it!

spbos:
Submitted a pull request.

PR #76, opened 2020-07-30
NoisyDetector subsampling enhancement 

spbos:
This commit adds subsampling capability to the NoisyDetector class. The subsampling factor and statistic parameters are added to __init__() to prevent unexpected behaviour. Also added test code. 

This enhancement was suggested in #75

### DIFF
--- a/hcipy/optics/detector.py
+++ b/hcipy/optics/detector.py
@@ -153,30 +153,30 @@ class NoisyDetector(Detector):
 			self._flat_field = flat_field
 
 	def integrate(self, wavefront, dt, weight=1):
-	'''Integrates the detector.
+		'''Integrates the detector.
 
-	The amount of power and dark current that the detector generates are calculated
-	for a given integration time and weight.
+		The amount of power and dark current that the detector generates are calculated
+		for a given integration time and weight.
 
-	Parameters
-	----------
-	wavefront : Wavefront or array_like
-		The wavefront sets the amount of power generated per unit time.
-	dt : scalar
-		The integration time in units of time.
-	weight : scalar
-		Weight of every unit of integration time.
+		Parameters
+		----------
+		wavefront : Wavefront or array_like
+			The wavefront sets the amount of power generated per unit time.
+		dt : scalar
+			The integration time in units of time.
+		weight : scalar
+			Weight of every unit of integration time.
 
-	'''
-	#The power that the detector detects during the integration.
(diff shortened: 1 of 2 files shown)


hcipy/optics/detector.py
```diff
@@ -130,6 +145,8 @@ def __init__(self, input_grid, dark_current_rate=0, read_noise=0, flat_field=0,
 		self.read_noise = read_noise
 		self.flat_field = flat_field
 		self.include_photon_noise = include_photon_noise
+		self.subsampling = subsampling
+		self.statistic = statistic
```
ehpor, line 149: You should make a distinction between `input_grid` (the grid on which the incoming wavefront should be defined) and `detector_grid` (the grid of the camera image). The input grid is now used as the detector grid. Not sure which one to use as the argument for `__init__()`. If input grid is used, you should check whether subsampling can be performed on this grid (in the initializer to catch mistakes early). Using detector grid as the initializer seems the right choice to me, but you should look at how that works in practice for a few minutes to make sure this is actually the better choice.

ehpor:
@spbos What's the progress on this? Any problems you encountered while addressing my comments?

spbos:
I have addressed the comments, but did not have time to push them yet. It will come soon.

spbos:
@ehpor I've addressed your comments a while back.

ehpor (approved):

ehpor:
I moved the subsampling functionality to the `Detector` base class to also have subsampling in the `NoiselessDetector` class. Still some leftover points, but I'm not gonna wait for those fixes before merging.

* Should the flatfield be applied before or after adding photon noise? Right now it's applied after, but I think it has to be before, acting as variations in the QE.
* QE is still missing (potentially a function of wavelength).
* Read noise should be quantized, right? Right now it isn't.
* I don't like how `weight` is used in the `Detector.integrate()`. Originally, it was intended to be used for polychromatic lightsources, where multiple wavelengths are sequentially integrated on the detector. In order to not add more dark noise than expected, you could add a weight of 1/N to each call of `Detector.integrate()`. This is fairly clunky and potentially wrongly implemented. We might have to sit down for this and discuss it further.
* `NoisyDetector` with default parameters now acts as a `NoiselessDetector` with photon noise, which can be turned off. It might be better to integrate both into a `SimpleDetector` class.

I'll merge once the CI finishes (excluding #79, which will be fixed in a separate PR).

PR #78, opened 2020-10-06
correction of a bug in get_strehl_from_focal

jmilou:
replacement of () by []


ehpor (approved):

ehpor:
CI doesn't finish because of an unrelated error (#79). I'll fix that later. Thanks for the bugfix, @jmilou!

PR #84, opened 2020-11-09
Update shack_hartmann_test.py

jmilou:
Hi Emiel, 

Thanks for adding the tutorial ! I quickly checked and had a couple of bugs that I fixed:
- make_focal_grid should be make_focal_grid_from_pupil_grid
- plot the absolute center of the SHWFS spot instead of the relative positions

I also replaced the dots by a cross to make it more visible on the plot.

### DIFF
--- a/examples/shack_hartmann_test.py
+++ b/examples/shack_hartmann_test.py
@@ -17,7 +17,7 @@ wavelength = 1e-6
 N = 256
 D = 0.01
 pupil_grid = make_pupil_grid(N, D)
-focal_grid = make_focal_grid(pupil_grid, 8, 20, wavelength)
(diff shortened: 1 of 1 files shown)


ehpor:
You modified the example, not the tutorial. The examples in `/examples` are very old code (mostly written before HCIPy became open source), and most do not run anymore without errors. I've been trying to migrate those to full tutorials instead of fixing them. The tutorial you and I wrote in #83 is in `/doc/tutorial_notebooks/ShackHartmannWFS`. You have to change to that branch and build the documentation yourself (https://docs.hcipy.org/0.3.1/development/project_setup.html#documentation). I didn't do the merge yet as I wanted to give you time to check it before then. If you want, I can do the merge now. Then you'll see the result appear on https://docs.hcipy.org/dev/tutorials/ once the build finishes (~20-30mins after the merge).

PR #87, opened 2020-11-16
Small fixes again

syhaffert:
Some small additions and bug fixes


hcipy/aperture/realistic.py
```diff
@@ -2,8 +2,63 @@
 from ..field import make_hexagonal_grid, Field
 from .generic import make_spider, circular_aperture, hexagonal_aperture, make_segmented_aperture, make_spider_infinite, make_obstructed_circular_aperture
 
-def make_vlt_aperture():
-	pass
+def make_vlt_aperture(normalized=False, with_spiders=True, orientation=0):
+	'''Make the VLT aperture.
+
```
ehpor, line 7: Can you add a source for the values used in this function? Citation or even webpage is fine, but I'd like to have at least some source listed here.

ehpor:
I agree with all changes here. After adding the source for the VLT aperture, I'll approve.

I reworked the way you did the supersampling of the z-tilt. This is how it's supposed to work (and already does for segmented apertures), although this is not yet consistent across HCIPy.

ehpor:
@dsdoelman @mkenworthy Apparently, the VLT aperture in this PR is based on your implementation. Do you have a source for the values for the angles and diameters used in here? Presumably ERIS documentation, but I'd like to be sure.

ehpor:
* The positioning of the spiders was incorrect in the proposed version. Spiders are linked to a point on the outer ring at a radius of 4.2197 meters, which is different (larger) than the pupil diameter of 4.0598 meters. This is described in VLT-SPE-AES-11310-0006. This problem is now fixed.
* I also added an optional cover for the M3 in stowed position (as implemented for ERIS).
* I also removed the option of orientation, as that's covered by a `make_rotated_aperture()` function. This function can also be applied to other apertures and field generators.

@dsdoelman @syhaffert @jmilou Would any of you be able to briefly check the code (in `/hcipy/aperture/realistic.py`) and generated aperture fits files (located in `/tests/baseline_for_apertures/vlt/*.fits.gz`)? I would like to have at least one extra pair of eyes look at this.

ehpor:
@dsdoelman pointed out a mistake in this implementation with the connection point of the spiders on the inside. I fixed this now. Here is a comparison, with overlay from the technical drawing.

hcipy_vlt_aperture

ehpor (approved):

jmilou:
Dear Emiel, 

I checked the files in /tests/baseline_for_apertures/vlt/*.fits.gz and your code to generate it. 
I think there are 3 problems: 

1. First as you see in the image you attached in your [previous comment](https://github.com/ehpor/hcipy/pull/87#issuecomment-729647585) with the technical drawing overlay, the M2 obscuration is larger than the actual M2 obscuration. 
The M2 diameter with the sky baffled retracted has a diameter of 1116mm as detailed in VLT-SPE-AES-11310-0006. This means that you should put:
`	central_obscuration_ratio = 1.116/pupil_diameter 
`
2. The pupil diameter is 8m, not 8.2m. 
8.2m is the M1 diameter which is larger than the pupil to give some margin when chopping is used in the IR. VLT-SPE-AES-11310-0006 says: "The pupil is determined by the secondary mirror which is designed to cover the 8 m aperture for the Nasmyth unvignetted field of view". So you should have: 
`	pupil_diameter = 8.0 # meter
`
3. The spider width is 5cm, not 4cm:
`	spider_width = 0.05 # 5cm meter
`

I did the change.

A minor comment concerning the pupil with the M3 cover opened to use the Cassegrain port. I guess this is to be used for ERIS simulation on UT4. UT4 has a different M2 compared to the other UT (deformable secondary mirror compared to the 3 other Dornier mirror). The obscuration is slightly larger than on the other UT and it has a fixed baffle (not retractable). However, despite a lot of efforts, I never managed to find technical drawings with the exact diameters of this baffle at ESO, so I am afraid this is an unknown. For this reason, I would put a warning when using this pupil for UT4.

Cheers
Julien

ehpor:
Great! Lots of new information to me. The implementation in here is based on ERIS (so UT4), specifically www.github.com/mkenworthy/eris_hci/tree/master/pupil.

1. Because this implementation was based on ERIS, that's like the result for the large M2 obscuration. The comment here is: `#DSM wind screen external diameter (3)`, where (3) refers to an ERIS memo (ERIS Memo No.: OAA-17-001, page 3/5). So I guess this is specific for UT4. UT1-3 should then have the smaller central obscuration.

2. Same is probably true here, though that makes less sense to me. The original comment here is `# Entrance pupil dia. (2)` where (2) refers to a different Eris document ("Table 4. Optical implementing BLF =500", VLT-TRE-ERI-14401-3103 page 8 of 14. ). Why this doesn't make sense to me is that ERIS on UT4 has a *larger* outer diameter (8.1x meter), and therefore *less* margin than other UTs? That feels weird, though it might be true.

3. It's nice to have an actual width for the spiders; the 4cm was an estimate according to the original implementation.

I personally did not know there was such a big (relatively speaking) difference between UTs. It might be best to have an option specifying which UT you want to make an aperture for.

jmilou:
Hi Emiel, 

OK I understand the difference if this is meant for UT4. For your information, all UTs were initially identical, and only the introduction of the DSM on UT4 changed this situation. 
2. I don't know if the pupil size on UT4 is now slightly larger (8.1m compared to 8.0m for the other UTs), but it could be that the tip/tilt amplitude of the DSM is smaller than the chopping amplitude of the original Dornier M2 on other UT, meaning that there was no point in not exploiting a larger pupil size. 
3. My mistake: the spiders are made of types of rectangular beams of width 39mm and 40mm respectively (see image) 
spider_section
spiders_overview

so your original implementation with 
`spider_width = 0.04 # 5cm meter`
was correct, sorry about that. 

The spiders have about the same width on UT4 compared to the other UTs, but they are higher because the DSM is fed with additional electric cables that are wrapped in a plastic cable wrap on top of the spiders. It does not change the pupil shape though.

Cheers

PR #94, opened 2021-02-17
phase_dot_diameter is in lam/D, so multiply by lam/D.

### LINKED ISSUE #93: Scaling of phase_dot_diameter for the ZernikeWavefrontSensor
On line 42 of zernike_wavefront_sensor.py, the phase_dot_diameter gets divided by lambda/D, but it should get multiplied by lambda/D, I think.

Edit: there might be something else going on, but at least in my personal simulation it seems to work with it flipped. To be investigated further.

@dsdoelman

ehpor:
Fixes #93. Also fixes the same bug in `VectorZernikeWavefrontSensorOptics`.

### DIFF
--- a/hcipy/wavefront_sensing/zernike_wavefront_sensor.py
+++ b/hcipy/wavefront_sensing/zernike_wavefront_sensor.py
@@ -39,7 +39,7 @@ class ZernikeWavefrontSensorOptics(WavefrontSensorOptics):
 		self.output_grid = input_grid
 
 		# Make the phase dot
-		phase_dot_diameter *= pupil_diameter / reference_wavelength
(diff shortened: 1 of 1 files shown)


ehpor:
Appveyor fails on Python 3.5 while importing numpy. As far as I can see, this is a CI issue, likely unrelated to us. I haven't found others with the same problem. We can merge after review.

syhaffert (approved):

syhaffert:
Here is the issue on the numpy github: https://github.com/numpy/numpy/issues/12977 
Seems to be a mixture of path/build errors.

ehpor:
Sure there are a bunch of issues on this, but anything specific to Appveyor.

PR #96, opened 2021-02-20
Fix return dtype of FTs for non-complex inputs

ehpor:


ehpor:
CI build issues. To be fixed later. Tests pass on my own machine.

Issue #112, opened 2021-10-22
Parallelizing pyramid modulation 


rmjc:
Hi all, 

I'm using HCIPy to simulate an adaptive optics system for the TMT, and I'm finding that by far the longest step in my closed-loop simulation is the pyramid modulation (specifically the for loop in the forward function of the ModulatedPyramidWavefrontSensorOptics class). It should be possible to parallelize the operations being done in the for loop, since there's no reason why each modulation step needs to be computed sequentially rather than in parallel. Shortening this step would make a big difference in making HCIPy practical for high time resolution simulations!

I tried to implement a parallelized version of the for loop using the multiprocessing package, but I'm having trouble doing so because python can't pickle PyramidWavefrontSensorOptics (I also get the error that the 'TipTiltMirror' object has no attribute 'is_regular' but I don't know what that means).  As a result, I'm finding that I need to re-create the PYWFS and TTM objects in each of the parallelized threads, which makes the speed improvement from parallelization much worse. 

If you folks can think of a practical way to make these objects picklable or otherwise parallelize / speed up the modulation, I would be very eager to use those capabilities. 

Cheers, 

Becky

syhaffert:
I am not sure if multi-processing will speed up the simulations. I have ran some tests by using multiprocess, but I saw that the single process approach was 3 to 4 times faster. In the figure below you can see the histograms
![image](https://user-images.githubusercontent.com/4592893/141411133-85c18dac-bbf0-4e23-b5b4-8289f443c85b.png)

What I also see in my CPU usage statistics is that the SP approach utilizes my cpus 100%, while the multiprocessing approach fluctuates between 70% and 98%. My guess is that most of the computation is done in the propagations, which are either FFTs or matrix-vector multiplications. Both are already very optimized to do the calculations efficiently with many cores. There may be a gain with the MP approach if you go to a cluster with many CPU cores. I only have 20 cores in my desktop, so I don't know how it scales to larger systems.

ehpor:
@syhaffert To confirm, multiprocess has no problem pickling these optical elements?

Also a general comment here. There usually is a minor gain in performance when going multiprocessing for parallelization, even in cases where a single process already uses the available CPU cores pretty effectively. Best performance then is often _not_ reached for N processes for N cores, but rather with only 2-4 processes.

[...]

PR #115, opened 2021-11-14
Autodetect FFTs for make_fourier_transform().

ehpor:
This PR adds a function `reconstruct_fft_grid_parameters()` which, given two grids, tries to reconstruct the parameters that went into `make_fft_grid()` to create the second out of the first. This is then used by the `make_fourier_transform()` function to produce an FFT in cases where the requested `output_grid` was actually a native FFT grid. Previously, this would create MFTs, which can be slower.

This also fixes a long-standing bug in hcipy, since open-source release in fact, where q factors that do not correspond to an integer number of zeropadding, would actually produce the output at different locations that the grid its saying it's calculating the FFT for. This case was not checked by the CI tests, and therefore slipped through our radar.

The speed improvements for the modulated PyWFS are ~25% for pupil grids of 128px across, and ~50% for pupil grids of 256px across (based on benchmarks on my own machine; timings might differ on yours).

@syhaffert I've had to fix a bug in how you were calculating the pupil diameter in `PyramidWavefrontSensor`. Please check if this is correct, and if this breaks any of your own code. It was a fencepost problem that you were trying to correct for, but that's incorrect in this case. Grids actually extend a little past their minimum and maximum coordinates, given that each pixel has a certain area.

This PR was brought about by a discussion in #112.


tests/test_fourier.py
```diff
@@ -71,12 +72,12 @@ def test_fourier_energy_conservation_2d():
...
-			for scale in [1,2]:
+		for shift_input in [[0,0], [0.1]]:
+			for scale in [1, 2]:
 				for shift_output in [[0,0], [0.1]]:
-					for q in [1,3,4]:
-						for fov in [1,0.5,0.8]:
-							for dims in [[8,8],[8,16],[9,9],[9,18]]:
+					for q in [1, 1.23, 3, 4]:
+						for fov in [1, 0.5, 0.8]:
+							for dims in [[8,8], [8,16], [9,9], [9,18]]:
```
syhaffert, line 80: Missing spaces in the list? [8,8] to [8, 8]?

ehpor: I tried to keep the pairs together, for visual reasons, but I'm fine with using spaces for all.

tests/test_fourier.py
```diff
@@ -95,16 +96,53 @@ def test_make_fourier_transform():
...
+
 	ft = make_fourier_transform(input_grid, q=1, fov=1, planner='measure')
 	ft = make_fourier_transform(input_grid, q=8, fov=0.1, planner='measure')
 
 	output_grid = CartesianGrid(UnstructuredCoords([np.random.randn(100), np.random.randn(100)]))
 	ft = make_fourier_transform(input_grid, output_grid)
 	assert type(ft) == NaiveFourierTransform
 
+def test_fft_grid_reconstruction():
+	for shift_input in [[0,0], [0.1]]:
```
syhaffert, line 118: spaces on lines 118, 120 and 123.

hcipy/fourier/fast_fourier_transform.py
```diff
@@ -39,10 +45,66 @@ def make_fft_grid(input_grid, q=1, fov=1):
 
 	delta = (2 * np.pi / (input_grid.delta * input_grid.dims)) / q
 	dims = (input_grid.dims * fov * q).astype('int')
-	zero = delta * (-dims / 2 + np.mod(dims, 2) * 0.5)
+	zero = delta * (-dims / 2 + np.mod(dims, 2) * 0.5) + shift
 
 	return CartesianGrid(RegularCoords(delta, dims, zero))
 
+def reconstruct_fft_grid_parameters(input_grid, fft_grid):
```
syhaffert, line 52: Does it make sense to name the grid to check fft_grid? Because its not necessarily an FFT grid. Its just a grid that is tested.

syhaffert (commented):
This looks good. I only had a couple formatting comments.

tests/test_fourier.py
ehpor: See previous comment. I tried to keep the pairs together, for visual reasons, but I'm fine with using spaces for all. Let me know which you think is cleaner.

tests/test_fourier.py
syhaffert: In a couple lines up there are spaces in between the pairs. I wasn't sure which convention was chosen.

hcipy/fourier/fast_fourier_transform.py
ehpor: `check_fft_grid()` feels to me like it should return either `True` or `False`, which is not what this function is doing / is supposed to be doing. But I strongly dislike the current name too. I'm open to suggestions. I tried `recover_fft_parameters()`, `reconstruct_fft_parameters()`, and `get_fft_grid_parameters()` and some variants. I don't like either of them.

tests/test_fourier.py
ehpor: I'll go full PEP8 then. Honestly I don't care that much for formatting of test code compared to library code, but I agree consistency is better. Same for below.

ehpor: Added spaces throughout the `test_optics.py` file.

ehpor:
@syhaffert You reviewed the changes in 1bb9b04 too? It's easy to overlook those two lines in this whole PR. Just making sure.

This is pending a fix for issue #116, and potentially a few lines of exception message clarification too.

hcipy/fourier/fast_fourier_transform.py
syhaffert: I think the function name is fine, but the parameter name fft_grid is a little bit confusing. I thought I would need to supply an fft grid to get the best matching fft grid parameters. Which is sort of true, because you only return thing if it actually is an fft grid. But it got me confused about what the function is trying to do.

Maybe the function should be split into two? A check if some grid is an FFT grid, and a recover parameters function? Because you now rely on the raising of ValueErrors for other functions. And I am not sure if that is the proper way of using warnings and errors.

syhaffert:
Yeah, I checked the lines in the PWFS. I think there is no issue with that change.

hcipy/fourier/fast_fourier_transform.py
ehpor: Gotcha, I misread that. The function I wrote assumes that `fft_grid` is a native FFT grid, and raises exceptions when this is not the case. Python uses exceptions way more liberal than you'd think (`IndexError`, `AttributeError`, `StopIteration`, etc..). Exceptions are not _that_ exceptional, compared to how they are supposed to be used in other languages.

Another reason for doing it this way is that while doing a check of whether a grid is a native FFT grid, you essentially have to do the exact same calculations as fully reconstructing the parameters (with the exception of `shift`, a single line of code).

Maybe add the following function?
```python
def is_fft_grid(input_grid, grid): # Maybe switch the arguments around? The same for the reconstruct... function?
    try:
        reconstruct_fft_grid_parameters(input_grid, grid)
    except ValueError:
        return False
    return True
```

The point is: `reconstruct_fft_grid_parameters()` (or whatever we end up calling it) will _need_ to raise an exception if the input grid is not a native FFT grid, as it has no way to signify this to the caller in any other way. Correct me if I'm wrong on that.

syhaffert: I think that is correct. You can't recover the parameters if the grid is not an fft grid. So raising a Value Error is the correct approach. I like the second function. I know it will add some overhead, but it's clearer what to expect back and how to check for an fft_grid.

ehpor:
This is now done. Tests pass on my machine. Tests pass on CI except for #116.

@syhaffert I'll do an admin merge after your approval.

syhaffert (approved):
I think it is good to go.

PR #118, opened 2021-11-15
Update fitsio

syhaffert:
I added the extension parameter to read_fits because I had to read in some files with multiple extensions.

### DIFF
--- a/hcipy/util/io.py
+++ b/hcipy/util/io.py
@@ -1,6 +1,6 @@
 import numpy as np
 
-def read_fits(filename):
+def read_fits(filename, extension=0):
 	'''Read an array from a fits file.
 
 	Parameters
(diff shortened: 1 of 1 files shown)


ehpor (approved):
I honestly see this function being removed sometime in the future as it is essentially a convenience method around an `astropy.io.fits` function. It was the counter function of `write_fits()` which allowed writing `Field`s to a fits file. But nowadays, there is the replacement of `read_field()` and `write_field()`, which also retain the grid of the field.

But this is a convenient feature to have.

ehpor:
Oh, and for the future, use _actual_ rebases (ie. `git rebase`), rather than merges for updating your branch with new updates on master.

syhaffert:
I think it's good to keep this function. I just went through the read_field function, and as far as I can see it can not handle non field data. For example, if I read in actual observations. It not necessary to have this function, but it makes it easier to work with hcipy in data reduction.

ehpor:
> I think it's good to keep this function. I just went through the read_field function, and as far as I can see it can not handle non field data. For example, if I read in actual observations. It not necessary to have this function, but it makes it easier to work with hcipy in data reduction.

Yes, it's a nice little convenience function - in fact, I use it all the time. But as it's not doing anything related to hcipy objects, people should preferably be using `astropy.io.fits` directly, or have this convenience function (or something similar) in their own code.

PR #120, opened 2021-11-16
Hale telescope pupil

syhaffert:
This PR adds the aperture for the 200-inch Hale Telescope at Palomar.


hcipy/aperture/__init__.py
```diff
@@ -15,10 +15,12 @@
 	'make_gmt_aperture',
 	'make_tmt_aperture',
 	'make_hexagonal_segmented_aperture',
+    'make_hexagonal_segmented_aperture',
```
ehpor, line 18: Why double this line?

hcipy/aperture/realistic.py
```diff
@@ -155,6 +155,51 @@ def func(grid):
...
+	spider_width = 2 * 0.024 # meter
+	central_obscuration_ratio = central_obscuration_diameter / pupil_diameter
+
+	if normalized:
+		spider_width /= pupil_diameter
+		pupil_diameter = 1.0
+
+	# Attachement points of the spiders at the central obscuration
+	box_heigth = 2 * 0.06
+	box_width = 2 * 0.0932 + central_obscuration_diameter
```
ehpor, line 189: Box width and height should be normalized as well. Put these where you define other parameters.

hcipy/aperture/realistic.py
```diff
@@ -155,6 +155,51 @@ def func(grid):
...
+		The Hale Telescope aperture.
+	'''
+	pupil_diameter = 5.08 # meter
+	central_obscuration_diameter = 1.86 # meter
+	spider_width = 2 * 0.024 # meter
+	central_obscuration_ratio = central_obscuration_diameter / pupil_diameter
+
+	if normalized:
+		spider_width /= pupil_diameter
+		pupil_diameter = 1.0
```
ehpor, line 185: You're not normalizing central obscuration diameter.

ehpor (changes requested):
Please look at _all_ test fits files. Those don't look correct.

hcipy/aperture/__init__.py
syhaffert: That happened during the rebasing. I removed it.

hcipy/aperture/realistic.py
syhaffert: Done.

hcipy/aperture/realistic.py
syhaffert: That's not necessary because I already calculate the central obscuration ratio a couple lines before this one.

ehpor: I always made the `if normalized` a kind of barrier, where no non-normalized values can pass through, even for values that are not used later on. That makes it harder to make normalization errors upon refactoring. But this'll do for now.

FYI: don't click resolved if you're not certain a comment is actually addressed. It's supposed to be the reviewer that checks and decides if all their comments are addressed sufficiently and click resolve if that is the case. It's not a checklist for you.

ehpor (approved):
LGTM.

PR #124, opened 2021-12-09
Switch to importlib from pkg_resources

ehpor:
`importlib` now supports loading from data files. This PR changes previous uses of `pkg_resources` to the new API. This reduces hcipy import times from 1.2sec to 0.7sec (on my machine).


ehpor:
This PR also:
* upgrades CI by removing source directory after installation and failing upon crashing tutorial notebooks
* updates documentation year
* removes requirements.txt
* fixes `fov` error message in `get_fft_parameters()`
* updates email, license files and supported Python versions in `setup.py`.

PR #126, opened 2022-05-27
Fix for missing asdf.fits_embed import.

ehpor:
Previously, `asdf.fits_embed` was available from `asdf` by mistake. Recently, this mistake in asdf was fixed, meaning that imports in hcipy need to be fixed as well.


ehpor:
Linter error fixed by #127 . Merging with admin privileges. The pickle bug is remaining.

PR #128, opened 2022-07-29
added CaF2 to the materials list

syhaffert:
see title


syhaffert:
This error seems to be related to the pickle issue @ehpor . Personally, I don't see any issue with this PR because I only added a single glass material to the materials list.

ehpor (approved):
Seems good.

PR #131, opened 2022-08-01
Bugfix for error message in `get_fft_parameters()`.

ehpor:
The variable `fov` was used before it was calculated.


ehpor:
Seems like #124 already did this fix. Closing this PR.

PR #135, opened 2022-08-02
Fix miscellaneous documentation warnings

ehpor:
* Default language set to English (en).
* Removed oversized underlined headings (Parameters and Returns) in docstrings.


syhaffert (approved):
looks good.

PR #136, opened 2022-08-03
Parametrized tests for Pytest

ehpor:
* This splits up large tests into multiple smaller ones.
* I also improved testing for apertures. Now, segmentation is tested for all configurations of the aperture, rather than only a single one. This showed that there was a bug in the HiCAT aperture code, which I fixed as well.
* Finally, the normalization function of apertures is tested now, meaning that unnormalized and normalized apertures should only have a scaling factor. This removes half of the reference aperture files as well.

All of this required recreating all reference apertures. While fits.gz files are deterministic, they include the filename inside the file contents, meaning that even a simple filename change results in a different binary blob. I manually checked all files to see if they were realistic, and I was super careful while changing over, doing the reference file change at the last minute.


ehpor:
> Merging [#136](https://codecov.io/gh/ehpor/hcipy/pull/136?src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Emiel+H.+Por) ([c9130f3](https://github.com/ehpor/hcipy/commit/c9130f3c5e6140d818562a08539c529e135e1525)) into [master](https://codecov.io/gh/ehpor/hcipy/commit/20f69b72673719d4c2bcced1a852577e274f7c6f?el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Emiel+H.+Por) ([20f69b7](https://github.com/ehpor/hcipy/commit/20f69b72673719d4c2bcced1a852577e274f7c6f)) will **decrease** coverage by `0.44%`.

Seems like I lost the LUVOIR-A Lyot stop tests along the way. I'll be re-adding those shortly.

ehpor:
@syhaffert This is now ready for review.

syhaffert (approved):
Looks good.

PR #137, opened 2022-08-04
Bugfix for missing shift during finding best FT method

ehpor:
Previously, the shift was ignored in the output grid of the `make_fourier_transform()` function. This PR includes the shift in the output grid which caused problems in a select case:

In cases where the possibility of an FFT was detected for the specified output grid, the output grid was replaced with an FFT grid. This FFT grid did not include the shift, which is strictly speaking not needed for calculating the performance of the FFT itself. The function would then do a performance comparison between FFT and MFT to select the fastest method. In cases where the MFT still was to be preferred, the MFT would be constructed using the FFT grid used for estimating the performance rather than the original requested output grid. The former didn't include the shift, which leads to an MFT being constructed for the wrong output grid.


syhaffert (approved):
No problems. Looks good.

PR #138, opened 2022-08-05
Add comparison (equality and inequality) operators to Grids.

ehpor:
This allows grids to be compared for equality:

```python
if grid1 == grid2:
    print('These two grids are equal.')
else:
    print('These two grids are different.')
```

The inequality operator (`!=`) is defined implicitly by Python 3 based on the equality operator.

Note that even if Grids contain the same coordinates but are different types, then the comparison will return False. This is intentional.

### DIFF
--- a/tests/test_field.py
+++ b/tests/test_field.py
@@ -200,6 +200,8 @@ def test_grid_hashing_and_comparison():
 	assert hash(grid1) != hash(grid3)
 	assert grid1 != grid3
 	assert grid3 != grid1
+	assert grid2 != grid3
+	assert grid3 != grid2
 
 	grid4 = make_pupil_grid(128)
 	print('start')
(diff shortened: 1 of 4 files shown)


tests/test_field.py
```diff
@@ -188,33 +188,53 @@ def test_field_svd():
...
 
 	grid2 = CartesianGrid(SeparatedCoords(copy.deepcopy(grid1.separated_coords)))
 	assert hash(grid1) != hash(grid2)
+	assert grid1 != grid2
+	assert grid2 != grid1
 
 	grid3 = CartesianGrid(UnstructuredCoords(copy.deepcopy(grid1.coords)))
 	assert hash(grid1) != hash(grid3)
+	assert grid1 != grid3
+	assert grid3 != grid1
```
syhaffert, line 203: Add an assert grid2 != grid3. Otherwise we will miss a code check for unstructure vs separated coords.

syhaffert (changes requested):
I have 1 small comment.

tests/test_field.py
ehpor: Done. Also added the reverse (which goes through a different code path.

syhaffert (approved):

ehpor:
Force merging because of the last commit (due to merges right before this one); the review is still valid.

PR #139, opened 2022-08-05
Add the JWST aperture.

ehpor:
This aperture is based on the one from WebbPSF. (FYI @mperrin)

This PR adds:
* Irregular polygons. I'm using Matplotlib Paths for evaluating those. This uses a bounding box, where only values inside the bounding box are evaluated. This provides an ~3x performance improvement. I didn't spend a lot of time on improving upon this even more.
* JWST aperture. This uses irregular polygons for the segments as they are slightly non-hexagonal. This is based on the pre-launch values as explained in the docstring/documentation.

This PR depends on #136, so test failures are expected until that PR is merged and this PR is rebased.

Reference apertures for test (with and without struts):

image


hcipy/aperture/generic.py
```diff
@@ -119,6 +121,37 @@ def func(grid):
...
+	vertices : iterable of length 2 iterables
+		The vertices of the polygon.
+
+	Returns
+	-------
+	Field generator
+		The generator for the polygonal aperture.
+	'''
+	p = Path(vertices)
+
```
syhaffert, line 138: Should we add a check here to see if it creates a valid polygon?

1. Is the path closed?
2. What happens when the vertices cross each other?

Or is this over kill at the moment? We could always add that in a future PR.

hcipy/aperture/realistic.py
```diff
@@ -1079,3 +1079,249 @@ def func(grid):
...
+			val /= pupil_diameter
+
+		pupil_diameter = 1
+
+	segments = [irregular_polygon_aperture(vertices) for vertices in segment_parameters.values()]
+	struts = [make_obstruction(irregular_polygon_aperture(vertices)) for vertices in strut_parameters.values()]
+
+	def func(grid):
+		ota = sum((segment(grid) for segment in segments))
+
```
syhaffert, line 1304: what is ota? It is probably better to use a descriptive variable name.

syhaffert (changes requested):
I have two comments for this PR.

hcipy/aperture/generic.py
ehpor: 1. The path is automatically closed by matplotlib.
2. For four vertices on a square, this is what happens if you have them in the right order:
image
and this for the wrong order:
image

hcipy/aperture/realistic.py
ehpor: optical telescope assembly. I agree with not having acronyms, so I'll change it.

hcipy/aperture/generic.py
syhaffert: Okay. Then this is resolved.

hcipy/aperture/realistic.py
syhaffert: ok.

syhaffert (approved):

PR #140, opened 2022-08-08
Improve aperture performance

ehpor:
This PR improves the speed of spiders and hexagonal segmented apertures by ~4x. This PR also adds tests for the circular and rectangular apertures.

Method
---
For spiders, I wrote specialized code for separated Cartesian grids, the most used grid in HCIPy. Most calculations can be performed much quicker because of the reduced array sizes,

For segmented apertures, I introduced a new keyword for Field generators; `return_with_mask`. This keyword, if it is `True` (default `False`), the function does not return a Field, but rather a `ndarray, mask` tuple. The mask can either be a tuple or a ndarray, to be used to mask the shaped or non-shaped version of a zero field with the original grid. This keyword allows for much more simplified additions to existing fields, which the `make_segmented_aperture()` function exploits to get higher performance.

This new keyword is not required for any Field generators, and is only used if it's available (as probed by the `inspect.signature()` function.

Benchmarks
---
Setup:
```python
pupil_grid = make_pupil_grid(512)

circle = circular_aperture(1)
rectangle = rectangular_aperture([1, 1])
hexagon = hexagonal_aperture(1)
spider = make_spider([0, 0.2], [0.5, 0.4], 0.03)
spider_infinite = make_spider_infinite([0, 0], 30, 0.03)

elt = make_elt_aperture(True)
gmt = make_gmt_aperture(True)
hicat = make_hicat_aperture(True)
tmt = make_tmt_aperture(True)
luvoir_a = make_luvoir_a_aperture(True)
luvoir_b = make_luvoir_b_aperture(True)
```
Current master branch:
```plain
HCIPy 0.4.1.dev196+geb74202 :
circle:     1.14 ms ± 11.6 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
rectangle:  591 µs ± 22.3 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
hexagon:    5.16 ms ± 145 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
spider:     9.01 ms ± 80.7 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
spider_inf: 5.68 ms ± 114 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
ELT:        604 ms ± 11.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
GMT:        178 ms ± 2.44 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
HiCAT:      84.6 ms ± 745 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
TMT:        392 ms ± 11.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
LUVOIR-A:   107 ms ± 902 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
LUVOIR-B:   42.1 ms ± 674 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
This PR:
```plain
HCIPy 0.4.1.dev200+g4395158 :
circle:     1.13 ms ± 10.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
rectangle:  583 µs ± 11.8 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
hexagon:    5.05 ms ± 162 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
spider:     1.87 ms ± 8.72 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
spider_inf: 1.78 ms ± 14.8 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
ELT:        150 ms ± 656 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
GMT:        178 ms ± 1.27 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
HiCAT:      28.6 ms ± 725 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
TMT:        103 ms ± 907 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
LUVOIR-A:   30.6 ms ± 341 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
LUVOIR-B:   12.4 ms ± 85.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```


tests/test_aperture.py
```diff
@@ -51,6 +52,25 @@ def test_regular_polygon_aperture():
...
 
+def test_circular_aperture():
+	pupil_diameter = 1
+
+	for diameter in [1, 0.5]:
+		name = 'circular/pupil'
+		name += '_small' if diameter < 0.7 else '_large'
+
+		# Use a functools.partial() here to avoid the name collision of the diameter argument.
+		check_aperture(functools.partial(circular_aperture, diameter=diameter), pupil_diameter, name)
```
syhaffert, line 63: circular_aperture will become make_circular_aperture due to PR #141 .

tests/test_aperture.py
```diff
@@ -51,6 +52,25 @@ def test_regular_polygon_aperture():
...
+
+def test_rectangular_aperture():
+	pupil_diameter = 1
+
+	for size in [[0.5, 0.5], [1, 0.5]]:
+		name = 'rectangular/pupil'
+		name += '_square' if size[0] == size[1] else '_elongated'
+
+		check_aperture(rectangular_aperture, pupil_diameter, name, size=size)
```
syhaffert, line 72: rectangular_aperture will become make_rectangular_aperture due to PR #141 .

tests/test_aperture.py
ehpor: I know, which is why #141 is a draft PR and will not be merged until this PR. Any merge conflicts will be fixed in #141, not the other way around.

tests/test_aperture.py
ehpor: I know, which is why #141 is a draft PR and will not be merged until this PR. Any merge conflicts will be fixed in #141, not the other way around.

hcipy/aperture/generic.py
```diff
@@ -302,13 +320,27 @@ def make_spider(p1, p2, spider_width):
...
+
+		x = x - shift[0]
+		y = y - shift[1]
 
 		x_new = x * np.cos(spider_angle) + y * np.sin(spider_angle)
 		y_new = y * np.cos(spider_angle) - x * np.sin(spider_angle)
-		spider_grid = CartesianGrid(UnstructuredCoords([x_new, y_new]))
 
-		return 1 - spider(spider_grid)
+		spider = x_new <= (spider_length / 2)
```
syhaffert, line 338: Wouldn't 
```
spider = abs(x_new) < spider_length / 2
spider *= abs(y_new) < spider_width / 2
```
be cleaner?

hcipy/aperture/generic.py
```diff
@@ -331,13 +363,26 @@ def make_spider_infinite(p, angle, spider_width):
...
+		y = y + p[1]
 
 		x_new = x * np.cos(spider_angle) + y * np.sin(spider_angle)
 		y_new = y * np.cos(spider_angle) - x * np.sin(spider_angle)
-		infinite_spider = np.logical_and(np.abs(y_new) <= (spider_width / 2), x_new >= 0)
 
-		return Field(1 - infinite_spider, grid)
+		infinite_spider = y_new <= (spider_width / 2)
+		infinite_spider *= y_new >= (-spider_width / 2)
+		infinite_spider *= x_new >= 0
```
syhaffert, line 383: `
spider = abs(y_new) <= spider_width / 2
`
?

ehpor:
For the spiders, doing comparison in this way is actually faster than an `abs()`:

```python
import numpy as np
import numexpr as ne

x = np.random.randn(1024*1024)

def f(x):
    res = x < 1
    res *= x > -1
    
    return res

%timeit f(x)

def g(x):
    res = np.abs(x) < 1
    
    return res

%timeit g(x)

def h(x):
    res = x * x < 1
    
    return res

%timeit h(x)

def i(x):
    res = ne.evaluate("x * x < 1", local_dict={'x': x})
    
    return res

%timeit i(x)
```

gives

```
1.77 ms ± 103 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
3.57 ms ± 85.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
3.59 ms ± 32.6 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
630 µs ± 23.5 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

So doing the comparisons separately is actually twice as fast as the abs. Using numexpr is even faster, but considering future changes to multi-backend, it's not worth having specialized code for numpy just yet. That'll be for the future once backends are separated.

hcipy/aperture/generic.py
ehpor: See https://github.com/ehpor/hcipy/pull/140#issuecomment-1219615161.

hcipy/aperture/generic.py
ehpor: See https://github.com/ehpor/hcipy/pull/140#issuecomment-1219615161.

syhaffert (approved):
Looks great. Let's merge.

ehpor:
> That is interesting. I wouldn't have thought that that would be the case. Okay no more comments from me then.

Yeah, me too at first. Something to do with the multiplication taking place in float64 vs boolean. I added a comment in the code with the counterintuitive result. I've used this trick too many times by now that it seemed obvious to me, but it's actually not trivial.

Merging after CI finishes.

PR #141, opened 2022-08-09
Rename existing aperture functions.

ehpor:
I deprecated the old aperture function names `circular_aperture()` and others, to unify the naming scheme between all aperture creation functions to `make_*_aperture()`.

The deprecated functions are still available, but will spit out a DeprecationWarning upon use. These decorators are now included for more general use in HCIPy as well, for future deprecations (instead of doing this manually, as I used to do).

This is a draft pull request because of this *question:* should we change this to `make_circle()` and similar names and drop the "aperture" as well?


hcipy/aperture/generic.py
```diff
@@ -515,3 +516,27 @@ def make_hexagonal_segmented_aperture(num_rings, segment_flat_to_flat, gap_size,
 		segment_positions = segment_positions.subset(mask)
 
 	return make_segmented_aperture(segment, segment_positions, return_segments=return_segments)
+
```
syhaffert, line 519: Do we want to make them deprecated immediately? Or what until the next major tick?

syhaffert (dismissed):

hcipy/aperture/generic.py
syhaffert: Never mind this comment. I read the new decorator code and I approve.

ehpor:
@syhaffert @dsdoelman You also agree on not renaming these `make_circle()` and similar names? I'd rather not go through name changes twice for no reason. (Tagging David, cuz he reminded/strongly suggested me to make this change, finally.) I'd like both of your responses before merging.

hcipy/aperture/generic.py
ehpor: I'm aiming to have them deprecated for at least one minor release *and* for at least a year. I unofficially do (and have done) this for all breaking changes.

dsdoelman:
I like the 'aperture'. I would never change functions like 'make_keck_aperture()'  -> 'make_keck()', so why change the circular aperture? I know people use it to create all kind of shapes for focal plane masks, but perhaps we can just make both and point them to the same function?

syhaffert:
I think we should keep aperture. This has been the style since the beginning. And in the focal plane a mask is still an aperture, just in a different plane. But we could always add another function for overloading if people think it is too much typing.

ehpor:
Okay, Thanks to you both! I'll keep these names then. It's easy to add aliases in the future if we desire that. Full name changes was what I was worried about. Merging this after #140.

syhaffert (approved):
Looks good to me.

ehpor:
Actually, I forgot to rebase to fix the deprecation warnings due to #140 (https://github.com/ehpor/hcipy/pull/140#discussion_r948010783 and https://github.com/ehpor/hcipy/pull/140#discussion_r948011154). Let me do that, and then I need another approval @syhaffert . Sorry for that.

ehpor:
Done. @syhaffert

syhaffert (approved):
looks good to merge.

PR #142, opened 2022-08-15
Feature/knife edge directions

syhaffert:
The Knife Edge coronagraph for multiple directions. A test has been added to check all directions and knife edge offsets.


syhaffert:
This branch is failing on the windows pickle bug. I will need to rebase this PR.

tests/test_coronagraphy.py
```diff
@@ -175,3 +175,33 @@ def test_lyot_coronagraph():
 	# Checks performance of the coronagraph and if the focal length does not introduce artifacts
 	assert (wf_foc.power.max() / norm) < 5e-3
 	np.testing.assert_allclose(wf_foc.power, wf_foc2.power)
+
+
+def test_knife_edge_coronagraph():
```
ehpor, line 180: This test passes, even when the `direction` parameter is completely ignored by the `KnifeEdgeLyotCoronagraph` class. We probably want to test that the E-field is the same but inverted for opposite-oriented knife edges.

tests/test_coronagraphy.py
```diff
@@ -175,3 +175,33 @@ def test_lyot_coronagraph():
...
+	prop = FraunhoferPropagator(grid, focal_grid)
+	lyot_stop = circular_aperture(0.95)(grid)
+
+	wf = Wavefront(aperture)
+	wf.total_power = 1.0
+	norm = prop(wf).power.max()
+
+	directions = ['+x', '-x', '+y', '-y']
+	for direction in directions:
+		knife_left = KnifeEdgeLyotCoronagraph(grid, direction=direction, apodizer=None, lyot_stop=lyot_stop)
```
ehpor, line 195: `knife_left`? Why the "left"?

tests/test_coronagraphy.py
```diff
@@ -175,3 +175,33 @@ def test_lyot_coronagraph():
...
+		wf_cor = prop(knife_left(wf))
+
+		assert (wf_cor.power.max() / norm) < 0.25
+
+	directions = ['+x', '-x', '+y', '-y']
+	knife_edge_shifts = [-1.0 * grid.x, 1.0 * grid.x, -1.0 * grid.y, 1.0 * grid.y]
+	for shift, direction in zip(knife_edge_shifts, directions):
+		pre_apodizer = np.exp(1j * 2 * np.pi * shift)
+		knife_left = KnifeEdgeLyotCoronagraph(grid, direction=direction, apodizer=pre_apodizer, lyot_stop=lyot_stop * np.conj(pre_apodizer))
```
ehpor, line 204: Same here. `knife_left`? Why the "left"?

tests/test_coronagraphy.py
```diff
@@ -175,3 +175,33 @@ def test_lyot_coronagraph():
...
+	aperture = circular_aperture(1)(grid)
+
+	focal_grid = make_focal_grid(q=5, num_airy=3)
+
+	prop = FraunhoferPropagator(grid, focal_grid)
+	lyot_stop = circular_aperture(0.95)(grid)
+
+	wf = Wavefront(aperture)
+	wf.total_power = 1.0
+	norm = prop(wf).power.max()
```
ehpor, line 191: Very minor: this should probably contain the Lyot stop as well. Totally unimportant for this test, but if you're making changes anyway, let's do this one as well.

ehpor (changes requested):
Some refinement of the test, but the rest looks good.

tests/test_coronagraphy.py
syhaffert: You want to normalize the PSF with respect to the Lyot Stop PSF?

ehpor: Of course, since that's what the planet looks like. And it's a coronagraphic image, not a PSF.

tests/test_coronagraphy.py
syhaffert: Added a test for both left/right and up/down symmetry.

tests/test_coronagraphy.py
syhaffert: This was a remnant of older code. I changed it now.

tests/test_coronagraphy.py
syhaffert: This was a remnant of older code. I changed it now.

tests/test_coronagraphy.py
syhaffert: Done

ehpor (approved):
LGTM.

PR #143, opened 2022-08-15
Fiber shift

syhaffert:
Added a keyword to the StepIndexFiber so that it can be placed at other positions than the origin.

### DIFF
--- a/hcipy/optics/fiber.py
+++ b/hcipy/optics/fiber.py
@@ -19,19 +19,19 @@ class StepIndexFiber(AgnosticOpticalElement):
 	fiber_length : scalar
 		The length of the optical fiber.
 	'''
-	def __init__(self, core_radius, NA, fiber_length, center=None):
+	def __init__(self, core_radius, NA, fiber_length, position=None):
 		self._core_radius = core_radius
 		self._NA = NA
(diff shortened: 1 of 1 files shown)


hcipy/optics/fiber.py
```diff
@@ -64,6 +66,17 @@ def core_radius(self, core_radius):
 		self._core_radius = core_radius
 		self.clear_cache()
 
+	@property
+	def position(self):
```
ehpor, line 70: Why mix `position` and `center`? I'd rather choose one or the other. My preference goes to `position`, but feel free to let me know any reason for the other name.

ehpor (changes requested):

hcipy/optics/fiber.py
syhaffert: I have adjusted the name. There was no particular reason why they were different.

ehpor:
Test failures are due to https://github.com/asdf-format/asdf/issues/1172. I'll restart CI once the fix comes out.

WilliamJamieson:
ASDF 2.11.2 and 2.12.1 have been released with jsonschema pins to resolve the issues.

ehpor (approved):
LGTM.

PR #145, opened 2022-08-17
VLTI aperture

syhaffert:
An implementation of the VLTI aperture. Based on the ESO P105 VLTI manual.


hcipy/aperture/realistic.py
```diff
@@ -100,6 +100,51 @@ def func(grid):
...
+		If this is True, the segments will also be returned as a ModeBasis.
+
+	Returns
+	-------
+	Field generator
+		The VLTI aperture.
+	segments : list of Field generators
+		The individual telescopes. Only returned when `return_segments` is True.
+	'''
+
```
ehpor, line 122: No empty line here.

hcipy/aperture/realistic.py
```diff
@@ -100,6 +100,51 @@ def func(grid):
...
+	baseline_UT13 = np.array([54.8, 86.5])
+	baseline_UT14 = np.array([113.2, 64.3])
+
+	relative_position = np.array([[0, 0], baseline_UT12, baseline_UT13, baseline_UT14])
+
+	# Calculate the middle between the extremes of the 4 telescopes
+	reference_position = (np.max(relative_position, axis=0) + np.min(relative_position, axis=0)) / 2
+
+	telescope_positions = np.array([-reference_position, baseline_UT12 - reference_position, baseline_UT13 - reference_position, baseline_UT14 - reference_position])
+	telescope_apertures = [make_shifted_aperture(make_vlt_aperture(telescope='ut{:d}'.format(i + 1), with_spiders=with_spiders), shift=telescope_positions[i]) for i in range(4)]
```
ehpor, line 134: Preferably use an f-string rather than a manual format. Also, maybe `enumerate()`?

ehpor (changes requested):
* Baselines vary as function of pointing on the sky. Currently, this function only returns the aperture for zenith pointing.
* Put pngs of the reference/baseline apertures as a comment.

syhaffert:
An image of the telescope positions
image

And the baselines
image

ehpor:
> * Put pngs of the reference/baseline apertures as a comment.

I meant `tests/baseline_for_apertures/vlti/pupil_without_spiders.fits.gz` and `tests/baseline_for_apertures/vlti/pupil.fits.gz` and the like. :) But these are great images for reference too!

ehpor (changes requested):
Looks much better. My two general comments remain. Also, the tests fail.

See https://github.com/ehpor/hcipy/pull/145#pullrequestreview-1076577979

ehpor:
> I think the aperture will also change with azimuth. Think about two telescopes, what happens when you observe at the horizon and parallel to the telescopes, you will only see a single aperture.

Which is why I said "pointing on the sky" rather than "altitude". 😛 

> 
> Should I also add the dOPD to the apertures ? Because the dOPD is Baseline * direction vector.

Maybe have a separate function for that rather than integrate it into `make_vlti_aperture()`? Or at least have a separate function for calculating the dOPDs? 

FYI, I would be fine with just having a few sentences in the docstring without any extra implementation for non-Zenith pointing, ie. leave it as it is now. Whichever you feel like doing. The failed tests need to be fixed though.

syhaffert:
Ok. I can implement both next week. I have already done the coding for the different on-sky pointings. I was thinking you were only talking about altitude because you mentioned zenith. My brain just completely ignored azimuth at that point.

syhaffert:
All changes have been implemented, tests have been added. All checks have been passed. I think this is ready to be merged.

ehpor:
You still didn't post the baseline apertures. It makes reviewing much easier. I'll try to get the review in tomorrow or the day after.

The baseline apertures used for the tests: from left to right, top to bottom, `pupil`, `pupil_non_zenith`, `pupil_non_zenith_without_spiders`, `pupil_without_spiders`:
image

ehpor (approved):
LGTM.

PR #149, opened 2022-08-18
Change interpolation default for `imshow_field()`

syhaffert:
Matplotlib had a small change in their default plotting behavior. It changed the default interpolation style to anti-aliased by using bilinear interpolation, see https://github.com/matplotlib/matplotlib/issues/17345 . The default keyword for imshow_field was changed to retain the default plotting style of hcipy.

### DIFF
--- a/hcipy/plotting/field.py
+++ b/hcipy/plotting/field.py
@@ -4,7 +4,7 @@ import numpy as np
 from ..field import Field
 
 def imshow_field(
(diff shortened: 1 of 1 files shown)


ehpor:
Indeed, this was changed in Matplotlib 3.2.0. See https://matplotlib.org/3.5.3/api/prev_api_changes/api_changes_3.2.0.html?highlight=image%20interpolation#default-image-interpolation

ehpor (approved):
Indeed, this behaviour is now the same as the docstring.

PR #150, opened 2022-08-18
Fix usage of deprecated function in tests.

ehpor:
Seems like I missed two instances in #141 due to post-merging #142.

Also fixes the deprecation warning message itself which didn't display the correct new function name due to mutable Python variables.

Also fixes two flake8 errors and modifies reviewdog to not filter out any flake8 errors (so that flake8 errors on lines that aren't modified by a PR are still included in the review).


syhaffert (approved):
If the PR passes all tests its ready for merging.

PR #156, opened 2022-09-06
Return mask on all exit paths.

ehpor:
On early exit of the `make_regular_polygon_aperture()`, a mask was never returned, even in cases where `return_with_mask` was `True`. This leads to an exception when a `make_segmented_aperture()` tries to unpack the result. This PR fixes this bug. The bug originated in #140.


syhaffert (approved):
Looks good to me.

PR #158, opened 2022-10-13
Add segment transmission option to ELT and TMT apertures

ivalaginja:
Minor change to have the same transmission options per segment like in some of the other segmented aperture functions.


ehpor (approved):
LGTM.

PR #161, opened 2022-10-18
added vacuum as glass material

syhaffert:
Created a material 'VACUUM' which has a refractive index of 1 for all wavelengths.


ehpor (approved):
LGTM. Whitespace is still difficult. 😛

PR #162, opened 2022-10-19
Faster modulated pyramid wavefront sensor

syhaffert:
This is the implementation of the faster modulated pwfs as mentioned in #112 . I have added a test to check the new method versus the old version and they agree quite well with each other. The maximum relative error is smaller than 2 percent, and the relative rms of the difference is ~0.2 percent.

### DIFF
--- a/hcipy/wavefront_sensing/pyramid.py
+++ b/hcipy/wavefront_sensing/pyramid.py
@@ -51,7 +51,7 @@ class ModulatedPyramidWavefrontSensorOptics(WavefrontSensorOptics):
 		self._pyramid_apodizers = []
 		for p in self.modulation_positions:
 			# The factor of two is to compensate for the factor of 1/2 of the mirror surface.
-			shifted_grid = self.focal_grid.shifted(2 * p) 
+			shifted_grid = self.focal_grid.shifted(2 * p)
 
 			pyramid_surface = evaluate_supersampled(surface_function, shifted_grid, 4)
 			pyramid_apodizer = SurfaceApodizer(Field(pyramid_surface, self.focal_grid), refractive_index)
--- a/hcipy/wavefront_sensing/pyramid.py
+++ b/hcipy/wavefront_sensing/pyramid.py
@@ -10,6 +10,11 @@ import numpy as np
 class ModulatedPyramidWavefrontSensorOptics(WavefrontSensorOptics):
 	'''The optical elements for a modulated pyramid wavefront sensor.
 
+	This class supports both a slow (default) and a fast modulation method, chosen by the
+	fast_modulation_method parameters. The fast method shifts the pyramid surface instead of
+	applying tilts to the incoming wavefront. While the fast modulation method can be up to
+	twice as fast, it needs to precompute all pyramid surfaces and will use up more memory.
+
 	Parameters
 	----------
 	pyramid_wavefront_sensor : WavefrontSensorOptics
@@ -20,8 +25,6 @@ class ModulatedPyramidWavefrontSensorOptics(WavefrontSensorOptics):
 		The number of steps per modulation cycle.
 	fast_modulation_method : boolean
 		If True the fast propagation method will be used. Default is False.
-		The fast method shifts the pyramid surface instead of shifting the PSF.
-		This reduces the number of propagations by approximately half.
 	'''
 	def __init__(self, pyramid_wavefront_sensor, modulation, num_steps=12, fast_modulation_method=False):
 		self.modulation = modulation
--- a/hcipy/wavefront_sensing/pyramid.py
+++ b/hcipy/wavefront_sensing/pyramid.py
@@ -20,6 +20,8 @@ class ModulatedPyramidWavefrontSensorOptics(WavefrontSensorOptics):
 		The number of steps per modulation cycle.
 	fast_modulation_method : boolean
 		If True the fast propagation method will be used. Default is False.
+		The fast method shifts the pyramid surface instead of shifting the PSF.
+		This reduces the number of propagations by approximately half.
 	'''
 	def __init__(self, pyramid_wavefront_sensor, modulation, num_steps=12, fast_modulation_method=False):
 		self.modulation = modulation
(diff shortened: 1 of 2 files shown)


hcipy/wavefront_sensing/pyramid.py
```diff
@@ -17,18 +18,43 @@ class ModulatedPyramidWavefrontSensorOptics(WavefrontSensorOptics):
 		The modulation radius in radians.
 	num_steps : int
 		The number of steps per modulation cycle.
+	fast_modulation_method : boolean
```
ehpor, line 21: You should explain what the (dis)advantages of the fast modulation method are.

hcipy/wavefront_sensing/pyramid.py
```diff
@@ -17,18 +18,43 @@ class ModulatedPyramidWavefrontSensorOptics(WavefrontSensorOptics):
...
+		separation = self.pyramid_wavefront_sensor._separation
+		refractive_index = self.pyramid_wavefront_sensor._refractive_index
+		wavelength_0 = self.pyramid_wavefront_sensor._wavelength_0
+
+		def surface_function(temp_grid):
+			return Field(-separation / (2 * (refractive_index(wavelength_0) - 1)) * (np.abs(temp_grid.x) + np.abs(temp_grid.y)), temp_grid)
+
+		self._pyramid_apodizers = []
+		for p in self.modulation_positions:
+			shifted_grid = self.focal_grid.shifted(2 * p)  # This factor two is to compensate for the 1/2 for the mirror surface.
```
ehpor, line 46: I prefer to have this inline comment on a separate line. Use inline comments sparingly.

hcipy/wavefront_sensing/pyramid.py
```diff
@@ -17,18 +18,43 @@ class ModulatedPyramidWavefrontSensorOptics(WavefrontSensorOptics):
...
+
+			pyramid_surface = evaluate_supersampled(surface_function, shifted_grid, 4)
+			pyramid_apodizer = SurfaceApodizer(Field(pyramid_surface, self.focal_grid), refractive_index)
+
+			self._pyramid_apodizers.append(pyramid_apodizer)
+
+		self.pupil_to_focal = self.pyramid_wavefront_sensor.pupil_to_focal
+		self.focal_to_pupil = self.pyramid_wavefront_sensor.focal_to_pupil
+
+		self._fast_modulation_method = fast_modulation_method
```
ehpor, line 56: Move this to the top of `__init__()` with all the other parameter-to-variable copies.

hcipy/wavefront_sensing/pyramid.py
```diff
@@ -42,13 +68,18 @@ def forward(self, wavefront):
...
 		'''
-		wf_modulated = []
+		if self._fast_modulation_method:
+			wf_foc = self.pyramid_wavefront_sensor.spatial_filter(self.pupil_to_focal(wavefront))
+			wf_modulated = [self.focal_to_pupil(apod(wf_foc)) for apod in self._pyramid_apodizers]
 
-		for point in self.modulation_positions.points:
-			self.tip_tilt_mirror.actuators = point
-			modulated_wavefront = self.tip_tilt_mirror.forward(wavefront)
+		else:
```
ehpor, line 75: No empty line before the else.

tests/test_wavefront_sensing.py
```diff
@@ -32,16 +32,27 @@ def test_modulated_pyramid_wavefront_sensor():
 
 	num_steps = 12
 
-	pywfs = PyramidWavefrontSensorOptics(pupil_grid, wfs_grid, 1.0)
+	pywfs = PyramidWavefrontSensorOptics(pupil_grid, wfs_grid, 1.0, q=3)
```
ehpor, line 35: Why this change? Because of systematic noise reasons of course, but does this mean that you _need_ to have higher `q` values for the fast modulation method? If so, that should definitely be in the docstring.

ehpor (changes requested):
Just simple style/docstring changes, plus one clarification.

tests/test_wavefront_sensing.py
syhaffert: The reason for this is that the code that chooses the correct focal plane sampling of the pyramid does not work correctly. I found that out last week. That is a bug that we need to fix. I am working on that in a separate issue. In the mean time I hard-coded the correct q for this simulation.

hcipy/wavefront_sensing/pyramid.py
syhaffert: done

hcipy/wavefront_sensing/pyramid.py
syhaffert: done

hcipy/wavefront_sensing/pyramid.py
syhaffert: done

hcipy/wavefront_sensing/pyramid.py
syhaffert: done

hcipy/wavefront_sensing/pyramid.py
ehpor: It's missing the disadvantages still. There need to be some, otherwise we would always use the fast method.

And I'd put this text in the main docstring rather than inside the parameters section.

Eg.:

```
'''The optical elements for a modulated pyramid wavefront sensor.

This class supports both a slow (default) and a fast modulation method, chosen by the
fast_modulation_method parameters. While the fast modulation method can be up to
twice as fast, it ....

Parameters
--------------
```

tests/test_wavefront_sensing.py
ehpor: Okay.

ehpor (changes requested):

hcipy/wavefront_sensing/pyramid.py
syhaffert: I think the only disadvantage is that it uses more memory. The two methods give really similar results so I don't know which one is actually more accurate compared to a true modulated pyramid. I went with:

`This class supports both a slow (default) and a fast modulation method, chosen by the
fast_modulation_method parameters. The fast method shifts the pyramid surface instead of
applying tilts to the incoming wavefront. While the fast modulation method can be up to
twice as fast, it needs to precompute all pyramid surfaces and will use up more memory.`

ehpor: That sounds good. I just wanted something as the disadvantage. Otherwise there is no reason to not remove the slow method altogether.

ehpor:
I'll approve once the tests have passed.

ehpor (approved):
LGTM.

PR #163, opened 2022-11-05
Emccd model

syhaffert:
I added functions to simulate the noise from the electron-multiplying gain process. This is a more accurate way to simulate the noise from EMCCD cameras. I followed the procedure from https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0053671 

Hirsch, Michael, et al. "A stochastic model for electron multiplication charge-coupled devices–from theory to practice." PloS one 8.1 (2013): e53671.


tests/test_util.py
```diff
@@ -83,3 +83,64 @@ def test_finite_difference():
...
+	emgain = 500
+
+	num_trials = 100
+	num_runs = 100000
+
+	noise = make_emccd_noise(photo_electron_flux * np.ones((num_trials, num_runs)), read_noise, emgain)
+	assert abs(np.sqrt(2 * photo_electron_flux) - np.std(noise / emgain)) / np.sqrt(2 * photo_electron_flux) < 1e-2
+
+	threshold = 8.0
+	assert np.max(abs(np.mean(noise, axis=0) / emgain - photo_electron_flux))  < threshold * np.sqrt(2 * photo_electron_flux)
```
ehpor, line 146: double space before <

tests/test_util.py
```diff
@@ -83,3 +83,64 @@ def test_finite_difference():
...
+	threshold = 8
+
+	num_trials = 100
+	num_runs = 100000
+	lam = 100.0
+	sigma = np.sqrt(lam / num_trials)
+
+	lam_realization = large_poisson(lam * np.ones((num_trials, num_runs)), thresh=1e6)
+
+	assert np.max(abs(np.mean(lam_realization, axis=0) - lam))  < threshold * sigma
```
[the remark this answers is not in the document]
ehpor: Also, testing the std instead of the max seems better to me. Same for the other distributions.

tests/test_util.py
```diff
@@ -83,3 +83,64 @@ def test_finite_difference():
...
+
+	num_trials = 100
+	num_runs = 100000
+	lam = 10000000.0
+	sigma = np.sqrt(lam / num_trials)
+
+	lam_realization = large_poisson(lam * np.ones((num_trials, num_runs)), thresh=1e6)
+
+	assert np.max(abs(np.mean(lam_realization, axis=0) - lam))  < threshold * sigma
```
ehpor, line 106: double space before <

hcipy/util/stats.py
```diff
@@ -32,3 +32,71 @@ def large_poisson(lam, thresh=1e6):
...
+		The electron multiplying gain of the EMCCD process. Must be larger than 0.
+
+	Returns
+	-------
+	array_like
+		The noisy realization of the EMCCD detection.
+	"""
+	photo_electrons = large_poisson(photo_electron_flux)
+
+	detector_counts = large_gamma(photo_electrons, emgain) + read_noise * np.random.standard_normal(photo_electron_flux.shape)
```
ehpor, line 100: This will give continuous values, not integers. Is this what you want? I'd expect this to give me ADU back.

ehpor (changes requested):
Some formatting, some clarifications about continuous vs discrete distributions.

hcipy/util/stats.py
```diff
@@ -32,3 +32,71 @@ def large_poisson(lam, thresh=1e6):
...
+	small = ~large
+
+	theta = theta * np.ones_like(lam)
+
+	# Use normal approximation if the shape parameter is large
+	n = np.zeros(lam.shape)
+	mean = lam * theta
+	std = np.sqrt(mean * theta)
+
+	n[large] = np.round(mean[large] + np.random.normal(size=np.sum(large)) * std[large])
```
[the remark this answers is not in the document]
syhaffert: Yeah I think I just copied it over from the large_poisson function. I will remove it to treat it similar to the small numbers.

hcipy/util/stats.py
syhaffert: This is the approximation that a lot of other people also use. And for large EMGAIN values you get hundreds of counts per photon. That will almost create a continuous distribution. If we really want to go to ADU then we need to implement the DAC layer too.

ehpor: Discussed with @syhaffert at OEI2023 (Lorentz Workshop in Leiden). We both agree now that there are better models out there which produce discrete distributions. But this simplified model is inherently continuous, as described in the paper, and is sufficiently close to the actual distribution that it warrants having it in. For more complicated models, we'll rely on external packages.

ehpor (approved):
LGTM after changes were made.

PR #165, opened 2022-11-11
Add FQPM coronagraph

### LINKED ISSUE #164: Add FQPM coronagraph
It looks like it would be very easy to add a FQPM coronagraph considering that it uses the same setup like `VortexCoronagraph()`. I have managed to do this locally, however, all I did was to replace the phase term inside of `VortexCoronagraph()` here with the appropriate phase for an FQPM:
https://github.com/ehpor/hcipy/blob/38b6e8974a3533430c7a7a7af86319c470ef078b/hcipy/coronagraphy/vortex.py#L68

If this is something that you'd want to have added, I presume a new class like `FourQuadrantPhaseMaskCoronagraph()` in a new module `fqpm.py` is the way to go, but do you have any advice on how to not duplicate an entire class? Do you have any ideas on how to split this to avoid that, or should I just go ahead and open a PR with an effectively copy-pasted class, just renamed (this really doesn't seem like a good idea). Maybe subclassing from something like a `PhaseMaskCoronagraph()`? Not sure how that would fly with the vector vortex (edit: I just saw that's just an agnostic optical element, so it's fine actually). I was tempted to just open a PR with what I have (copied code...) but I wanted to avoid a mess.

There is also some fiddling with the focal grid that I still need to do to clean it up and make it work properly, but overall this doesn't look like that much work.

ivalaginja:
Fixes #164.

I have implemented the restructuring suggested by @syhaffert in #164. Notes:
- ~~I replaced `np.sign()` in creating the FQPM with a subtraction of two orthogonal heaviside functions which gives a better attenuation.~~
- The location of the new classes was chosen by me arbitrarily and I'll be happy to change that.
- ~~The default parameters for the `VortexCoronagraph` don't work as well for the `FQPMCoronagraph`, so I adapted them. There might still be parameter combinations that work better than what I put in here.~~
- ~~The `FQPMCoronagraph` needs a focal-plane grid that does not contain the origin (centered between pixels instead of on a pixel) to work well. I have added this tentatively in commit [bf12a81](https://github.com/ehpor/hcipy/pull/165/commits/bf12a814f265ce3b372e6be97d5f98c9f60e6959) but this breaks the `VortexCoronagraph`. Waiting on some advice from @syhaffert to find a solution for flexible focal-plane grids - not sure if a simple if statement inside of `MultiScaleCoronagraph` would suffice... for example to test a hard-coded boolean attribute inside of the subclasses? Something like `self.grid_contains_origin = True/False`?~~
- ~~Missing tests, missing documentation.~~

### DIFF
--- a/hcipy/coronagraphy/multi_scale.py
+++ b/hcipy/coronagraphy/multi_scale.py
@@ -3,7 +3,7 @@ from scipy.signal import windows
 
 from ..optics import OpticalElement, Apodizer, Wavefront
 from ..propagation import FraunhoferPropagator
-from ..field import make_focal_grid, Field
+from ..field import make_focal_grid, make_pupil_grid, Field
 from ..aperture import make_circular_aperture
 from ..fourier import FastFourierTransform, MatrixFourierTransform, FourierFilter
 
@@ -64,7 +64,8 @@ class MultiScaleCoronagraph(OpticalElement):
 			q = qs[i]
 			num_airy = num_airys[i]
 
-			focal_grid = make_focal_grid(q, num_airy, pupil_diameter=pupil_diameter, reference_wavelength=1, focal_length=1)
+			#focal_grid = make_focal_grid(q, num_airy, pupil_diameter=pupil_diameter, reference_wavelength=1, focal_length=1)
+			focal_grid = make_pupil_grid(2 * q * num_airy, 1)
 			focal_mask = phase_pattern(focal_grid)
 
 			focal_mask *= 1 - make_circular_aperture(1e-9)(focal_grid)
--- a/hcipy/coronagraphy/multi_scale.py
+++ b/hcipy/coronagraphy/multi_scale.py
@@ -1,0 +1,161 @@
+import numpy as np
+from scipy.signal import windows
+
+from ..optics import OpticalElement, Apodizer, Wavefront
+from ..propagation import FraunhoferPropagator
+from ..field import make_focal_grid, Field
+from ..aperture import make_circular_aperture
+from ..fourier import FastFourierTransform, MatrixFourierTransform, FourierFilter
+
+class MultiScaleCoronagraph(OpticalElement):
+	'''A phase mask coronagraph.
+
+	This :class:`OpticalElement` simulates the propagation of light through
+	a phase mask in the focal plane. To resolve the singularity of the
+	phase plate, a multi-scale approach is made. Discretisation errors made at
+	a certain level are corrected by the next level with finer sampling.
+
+	Parameters
+	----------
+	input_grid : Grid
+		The grid on which the incoming wavefront is defined.
+	phase_pattern : field_generator
+		The focal plane phase mask.
+	lyot_stop : Field or OpticalElement
+		The Lyot stop for the coronagraph. If it's a Field, it is converted to an
+		OpticalElement for convenience. If this is None (default), then no Lyot stop is used.
+	q : scalar
+		The minimum number of pixels per lambda/D. The number of levels in the multi-scale
+		Fourier transforms will be chosen to reach at least this number of samples. The required
+		q for a high-accuracy vortex coronagraph depends on the charge of the vortex. For charge 2,
+		this can be as low as 32, but for charge 8 you need ~1024. Lower values give higher performance
+		as a smaller number of levels is needed, but increases the sampling errors near the singularity.
+		Charges not divisible by four require a much lower q. The default (q=1024) is conservative in
+		most cases.
+	scaling_factor : scalar
+		The fractional increase in spatial frequency sampling per level. Larger scaling factors
+		require a smaller number of levels, but each level requires a slower Fourier transform.
+		Factors of 2 or 4 usually perform the best.
+	window_size : integer
+		The size of the next level in the number of pixels on the current layer. Lowering this
+		increases performance in exchange for accuracy. Values smaller than 4-8 are not recommended.
+	'''
+	def __init__(self, input_grid, phase_pattern, lyot_stop=None, q=1024, scaling_factor=4, window_size=32):
+		self.input_grid = input_grid
+		pupil_diameter = input_grid.shape * input_grid.delta
+
+		if hasattr(lyot_stop, 'forward') or lyot_stop is None:
+			self.lyot_stop = lyot_stop
+		else:
+			self.lyot_stop = Apodizer(lyot_stop)
+
+		levels = int(np.ceil(np.log(q / 2) / np.log(scaling_factor))) + 1
+		qs = [2 * scaling_factor**i for i in range(levels)]
+		num_airys = [input_grid.shape / 2]
+
+		focal_grids = []
+		self.focal_masks = []
+		self.props = []
+
+		for i in range(1, levels):
+			num_airys.append(num_airys[i - 1] * window_size / (2 * qs[i - 1] * num_airys[i - 1]))
+
+		for i in range(levels):
+			q = qs[i]
+			num_airy = num_airys[i]
+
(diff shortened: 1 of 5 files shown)


ivalaginja:
Using a test like already exists for the VortexCoronagraph, we can see that this FQPM doesn't reach quite the same result. Below the test, printing the resulting values instead of using assertions:
```python
def test_fqpm_coronagraph():
	pupil_grid = make_pupil_grid(256)
	focal_grid = make_focal_grid(4, 32)

	aperture = make_circular_aperture(1)
	aperture = evaluate_supersampled(aperture, pupil_grid, 8)

	lyot = make_circular_aperture(0.99)
	lyot = evaluate_supersampled(lyot, pupil_grid, 8) > 1 - 1e-5

	fqpm = FQPMCoronagraph(pupil_grid, q=4, scaling_factor=2, window_size=256)

	wf = Wavefront(aperture)
	wf.total_power = 1

	img_ref = prop(wf)

	wf = fqpm(wf)
	wf.electric_field *= lyot
	img = prop(wf)

        # assert img.total_power < 1e-6
        # assert img.intensity.max() / img_ref.intensity.max() < 1e-8

	print(img.total_power)
	print(img.intensity.max() / img_ref.intensity.max())

test_fqpm_coronagraph()
```

And the output:
```bash
1.3609898015020445e-05
1.8636041696954595e-07
```

ivalaginja:
@syhaffert any thoughts on this?

syhaffert:
I have been running through some tests and I think for the FQPM the high resolution is necessary along the edges of the focal plane mask. Not only in the middle. The multiscalecoronagraph zooms in at the center of the mask, exactly where the singularity of the vortex is. For the FQPM we should zoom in along the edges. I have made something that might work, but I still need to figure out what the correct parameters are. Its based on the KnifeEdgeCoronagraph.

ivalaginja:
Any news @syhaffert ? (This is not pressing in any way, just checking in.)

syhaffert:
Not yet. I still have to resolve a pixel offset somewhere in the code. Very frustrating last bug.

ivalaginja:
Any luck with that bug?

ivalaginja:
Happy new year! Any news on this?

ivalaginja:
Motivate by last week's workshop, I just wanted to up this here again ^

syhaffert:
@ehpor and me discussed a better implementation last week using adaptive sampling and multi-scale transforms. I am now on an observing run and I don't know if I have time to implement it. But usually during some slow observing nights I start a lot of hcipy coding, so I might have a first go at it.

ivalaginja:
Any news on this? Pierre got curious about the FQPM in hcipy in light of having a new intern and also looking ahead to the CDI workshop in June.

ivalaginja:
General update:

@ehpor helped me figure out how to feed a FQPM phase mask properly into the multi-scale class. I adapted the code accordingly and I think for the matter of adding a FQPM to hcipy, this is good to go. If there is ever the need to adapt the multi-scaling to the discontinuities of the FQPM in particular, this can be done separately. I don't think anybody needs this right at this moment.

I have double-checked that the `VortexCoronagraph` gives exactly the same result before and after this PR, the Lyot plane array and the final focal plane array subtract to zero exactly.

The FQPM propagation with the defaults from this PR look as follows (note the large range of lam/D in the focal plane for the display):
Screenshot 2023-06-27 at 16 14 46

Code this was created with:
```python
import hcipy


pup_size_px = 512    # [px]
pup_size_m = 2.0   # [m]
pupil_grid = hcipy.make_pupil_grid(pup_size_px, pup_size_m)

aperture = hcipy.evaluate_supersampled(hcipy.circular_aperture(1), pupil_grid, 32)
ls_undersizing = 0.89
lyot_mask = hcipy.circular_aperture(ls_undersizing)(pupil_grid)
lyot_stop = hcipy.Apodizer(lyot_mask)

samp = 8    # px per lam/D
imglamD = 200    # focal-plane images half size in units of lam/D
focal_px = samp * imglamD * 2
focal_grid_fpm = hcipy.make_uniform_grid(dims=(focal_px, focal_px), extent=imglamD * 2)
prop = hcipy.FraunhoferPropagator(pupil_grid, focal_grid_fpm, focal_length=1)

coro = hcipy.FQPMCoronagraph(pupil_grid)

wf = hcipy.Wavefront(aperture)
lyot_plane = coro(wf)
post_lyot_mask = lyot_stop(lyot_plane)
img = prop(post_lyot_mask).intensity

plt.figure(figsize=(18, 9))
plt.subplot(1, 2, 1)
hcipy.imshow_field(lyot_plane.intensity, cmap='inferno', norm=LogNorm())
plt.title('Lyot plane')
plt.colorbar()
plt.subplot(1, 2, 2)
hcipy.imshow_field(np.log10(img), cmap='inferno')
plt.title('Coro PSF')
plt.colorbar()
```

ivalaginja:
Ready for review @syhaffert  - just fyi @ehpor in case you had anything to add, since you talked both to Sebastiaan and me about it.

ivalaginja:
@syhaffert @ehpor anybody habe some time for a review? :)

syhaffert (commented):
Sorry for the late review. I think this is pretty much what I coded up too a couple of months ago. So I think it's a good approach. The only comment I have is that the documentation for the MultiScaleCoronagraph still explicitly mentions the vortex coronagraph.

hcipy/coronagraphy/multi_scale.py
```diff
@@ -0,0 +1,158 @@
...
+        The grid on which the incoming wavefront is defined.
+    complex_mask : field_generator
+        The complex focal-plane phase mask.
+    lyot_stop : Field or OpticalElement
+        The Lyot stop for the coronagraph. If it's a Field, it is converted to an
+        OpticalElement for convenience. If this is None (default), then no Lyot stop is used.
+    q : scalar
+        The minimum number of pixels per lambda/D. The number of levels in the multi-scale
+        Fourier transforms will be chosen to reach at least this number of samples. The required
+        q for a high-accuracy vortex coronagraph for example depends on the charge of the vortex. For charge 2,
```
ivalaginja, line 29: @syhaffert This is one of the instances where the vortex is still explicitly stated in the docstring of the now generalized multi-scale coronagraph. I see this as an example to illustrate the parameter `q` though, so I purposefully left it there. Or do you think it would be better in the Vortex child class? I am not a fan of that since `q` is a parameter of this class here, but of course it depends on the actual coronagraph it's used with.

hcipy/coronagraphy/multi_scale.py
```diff
@@ -0,0 +1,158 @@
...
+        if self.lyot_stop is not None:
+            lyot = self.lyot_stop.forward(lyot)
+
+        return lyot
+
+    def backward(self, wavefront):
+        '''Propagate backwards through the coronagraph.
+
+        For a vortex coronagraph, this essentially a forward propagation through the same
+        coronagraph, but with the sign of its charge flipped.
```
ivalaginja, line 129: @syhaffert same here, except I find it less critical here so I am happy to move it or delete it. Please let me know!

hcipy/coronagraphy/multi_scale.py
syhaffert: I think this is okay. It helps people understand what the correct order is for the parameter.

syhaffert (commented):
I think everything looks good. Please address the documentation of the backward propagation. After that it is ready for merging.

ivalaginja:
Sounds good, and all done!

ivalaginja:
We're so close ^^ Anything else on this @syhaffert or are we good to go? :)

syhaffert (approved):
This is ready for merging.

PR #166, opened 2022-11-23
Very small bug fix in Wavefront class

syhaffert:
I found a code breaking typo in the Wavefront class. The wavenumber assignment method used npf.pi instead of np.pi. I fixed this.


ehpor (approved):
idk how this one slipped through for this long 😆

syhaffert:
probably because nobody uses wavenumbers. at least I know I never use it.

PR #168, opened 2022-11-26
fix bug in hexgrid and documentation

syhaffert:
I have cleared up the documentation of make_hex_grid and I have switched some internal variable names to use the actual correct geometric properties.

I have run a couple of tests with the TMT and ELT aperture to make sure they still create identical apertures and those tests passed.


syhaffert:
@ehpor I think we can actually close this PR because we decided that it was not an implementation error.

ehpor:
@syhaffert Agreed. I might make a separate PR clarifying the language a bit, but that's a separate PR.

PR #170, opened 2023-02-23
Fix bug in interpolating big-endian ndarray read from fits file.

ehpor:
This PR fixes a bug in the Xinetics influence functions where the influence function file was read into a big-endian array, and the interpolation function doesn't support that anymore. The quick fix is to cast to a little-endian array before doing any interpolation. This is a minimal computational cost.


ivalaginja (approved):
Tests are passing, I am happy it was this easy of a fix.

PR #171, opened 2023-02-25
Add Keck Lyot stop in L band

ivalaginja:
Move to here from #155 since @vkooten and I weren't sure about how exactly to deal with some details of the Lyot stop and we want to merge the Keck aperture asap.


ehpor:
@vkooten @ivalaginja What's the progress on this? Have those details been figured out?

ivalaginja:
Unfortunately I believe this is a dead-end PR without a plan forward. We could reach out to @vkooten offline to check with her, but in my opinion this can be closed without merging.

ehpor:
@ivalaginja Reached out to @vkooten on a hopefully current email.

PR #172, opened 2023-03-01
Add missing colon for conda installation

ivalaginja:
Conda channels need two colons when installing a package from them.


ehpor (approved):
Thanks for the fix. Discovered during the Lorentz Workshop, but I don't remember if that was you or someone else.

ivalaginja:
It was Roser!

PR #175, opened 2023-03-28
Add problem matcher to annotate flake8 errors in PRs.

### LINKED ISSUE #174: Reviewdog doesn't appear to work on forks
Output from a CI:
```
[action-flake8] Checking python code with the flake8 linter and reviewdog...
reviewdog: This GitHub token doesn't have write permission of Review API [1], 
reviewdog: Reporting results for "flake8"
```

A solution is to convert over a manual annotation script.

ehpor:
Fixes #174 by removing reviewdog as a dependency.

### DIFF
--- a/.github/flake8_problem_matcher.json
+++ b/.github/flake8_problem_matcher.json
@@ -1,0 +1,32 @@
+{
+    "problemMatcher": [
+        {
+            "owner": "flake8-error",
+            "severity": "error",
+            "pattern": [
+                {
+                "regexp": "^([^:]*):(\\d+):(\\d+): ([E]\\d\\d\\d) (.*)$",
+                "file": 1,
+                "line": 2,
+                "column": 3,
+                "code": 4,
+                "message": 5
+                }
+            ]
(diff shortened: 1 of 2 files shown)


ivalaginja:
Is this supposed to already fix things for the tests on this branch? Asking because they're still waiting for the flake8 status to be reported after all tests finished and passed.

ehpor:
> Is this supposed to already fix things for the tests on this branch? Asking because they're still waiting for the flake8 status to be reported after all tests finished and passed.

I think that `flake8` requirements is for reviewdog, which is being disabled by this PR. If so, I should make it optional. I'll look into that now.

FYI, this is a copy of the `hicat-package2` problem matcher.

ehpor:
Unconfirmed, since Github doesn't tell me the source other than "Github Action", but likely that came from reviewdog. I removed it anyway, since the Lint action takes care of that now.

ivalaginja (approved):
Ok - looks good to me in that case.

PR #176, opened 2023-04-03
Dispersion optics

syhaffert:
I have added some optics that introduce dispersion (prism and grating) and some helper functions. These will be useful for simulating spectrographs and ADCs. I have tried to make comprehensive tests and document all functionality.

### DIFF
--- a/hcipy/optics/dispersion_optics.py
+++ b/hcipy/optics/dispersion_optics.py
@@ -61,7 +61,7 @@ class TiltElement(SurfaceApodizer):
 	refractive_index : scalar or function
 		The refractive index of the material. The default is 2.0 which makes it achromatic and exact.
 	'''
-	def __init__(self, angle, orientation = 0, refractive_index = 2.0):
+	def __init__(self, angle, orientation=0, refractive_index=2.0):
 		self._angle = angle
 		self._orientation = orientation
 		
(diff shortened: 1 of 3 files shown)


ehpor (changes requested):
Overall looks like a worthwhile change and definitely something that I'll use myself instead of the few rudimentary classes that I built personally and never bothered to clean up.

Some PEP8 stylistic changes regarding assignment of lambdas to variables. These should be replaced by a function definition. There are a few more than I indicated, so look yourself at other places where you made the same mistake. Also some missing units in the docstrings (also way more than I indicated; just look through all the places where there is an angle and add "radians").

syhaffert:
I have addressed your comments and made some changes. I think it should be ready now.

hcipy/optics/dispersion_optics.py
```diff
@@ -0,0 +1,325 @@
...
+		'''
+		return Field(grid.rotated(self._orientation).y * np.tan(self._angle), grid)
+
+class ThinPrism(TiltElement):
+	'''A thin prism that operates in the paraxial regime.
+
+	Parameters
+	----------
+	angle : scalar
+		The wedge angle of the prism.
```
ehpor, line 131: Still a missing unit here.

ehpor (changes requested):
The rest looks good.

ehpor (approved):
Thanks for all these new optical elements!

PR #178, opened 2023-04-06
Zoom Fast Fourier Transform and Chirp Z-Transform

ehpor:
The ZoomFFT is a Fourier transform based on the chirp Z-transform. Asymptotically it is faster than the MFT - since it's implemented using FFTs, its complexity is O(N^2 log(N)) for 2D transforms - the point of equality is only reached at array sizes of 1k x 1k (input) to 1k x 1k (output) or larger.

This Fourier transform will be added to the `make_fourier_transform()` function with a later PR. This PR is just for adding its implementation and the tests to support it long term.

One short-term caveat for the tests: the chirp Z-transform is tested by comparing with the scipy implementation of the CZT. We're not using that implementation since the FFTs that it uses are not accelerated. However, the scipy implementation is only available from version 1.8.0 onwards, which does not support Python 3.7. Therefore, I had to disable the CZT tests when scipy >=1.8.0 is not installed, so only on Python 3.8+. This caveat will be resolved when support for Python 3.7 will be dropped from HCIPy.

### DIFF
--- a/hcipy/fourier/chirp_z_transform.py
+++ b/hcipy/fourier/chirp_z_transform.py
@@ -1,0 +1,84 @@
+import numpy as np
+from scipy.fft import next_fast_len
+
+try:
+	import mkl_fft._numpy_fft as _fft_module
+except ImportError:
+	_fft_module = np.fft
+
+class ChirpZTransform:
+	'''The Chirp Z-transform (CZT).
+
+	This class evaluates:
+
+	.. math:: X_k = \sum_{n=0}^{N-1} x(n) z_k^{-n}
+
+	where :math:`N` is the number of points along the real space, :math:`M`
+	is the number of points in the Z-plane. The Chirp Z-transform samples
+	the Z-plane on equidistant points along spiral arcs, where the arcs are
+	given by the starting point :math:`A` and the complex ratio between points
+	:math:`W`:
+
+	.. math:: z_k = A * W^{-k}, k=0, 1, \ldots, M - 1
+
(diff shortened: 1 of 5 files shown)


ehpor:
@syhaffert This is still waiting for your review, and is now blocking another PR that I'm working on.

hcipy/fourier/chirp_z_transform.py
```diff
@@ -0,0 +1,109 @@
...
+		overflow.
+	a : scalar
+		The starting point in the complex Z-plane. Note: ensure that its absolute
+		value is close to one, so that a**max(n, m) does not overflow.
+	'''
+	def __init__(self, n, m, w, a):
+		# Ensure that w and a are complex scalars.
+		w = complex(w)
+		a = complex(a)
+
```
syhaffert, line 47: Do we need to add catch values that will cause an overflow error?

syhaffert (commented):
Looks good. I had 1 comment.

hcipy/fourier/chirp_z_transform.py
ehpor: This is not a trivial check, and imo would be part of the responsibility of the user. Any reasonable implementation that uses chirp Z-transforms would not even come close to overflowing/underflowing the spiral in the Z-plane.

syhaffert: Okay. I thought there might be a simple check. I looked through the source code of scipy and they also don't do a check. This issue is resolved.

syhaffert (approved):
This PR is approved for merging.

PR #179, opened 2023-04-08
Upgrade packaging to `pyproject.toml`.

ehpor:
Modern Python packages can contain a pyproject.toml file, first introduced in [PEP 518](https://peps.python.org/pep-0518/) and later expanded in [PEP 517](https://peps.python.org/pep-0517/), [PEP 621](https://peps.python.org/pep-0621/) and [PEP 660](https://peps.python.org/pep-0660/). This file contains build system requirements and information, which are used by pip to build the package.

As `setup.py` has been deprecated, we need to move to `pyproject.toml` packaging at some point. This PR accomplishes this. This also updates the documentation accordingly.


syhaffert (approved):
Looks good to me. This can be merged.

PR #180, opened 2023-04-08
Fourier shift, shear and rotation operators.

ehpor:
## Description

This pull request introduces three new classes for performing image operations in the Fourier domain: shift, shear and rotation. These operators are implemented as classes that can be instantiated and then applied to Field objects.

## Added classes

* `FourierShift`: Shifts an image by a given amount. This is implemented by applying a phase ramp in the Fourier domain. The Fourier transform is only calculated along that axis.
* `FourierShear`: Shears an image along a specified dimension. The Field is only Fourier-transformed along that axis, saving computation time.
* `FourierRotation`: Rotates an image by a given angle. This is implemented using the three-shear method. For rotation angles larger than 45 degrees, the operator breaks the operation into multiple, smaller rotations to maintain correctness. Automatic zero-padding is used to prevent boundary artifacts during the shearing process.

## Added tests

Unit tests have been added for all three new operators to ensure their correctness and adjointability. These tests cover basic functionality, adjoint operations, and edge cases.


ehpor:
@syhaffert Tests are failing on the actuator positions.

syhaffert (commented):
Should we only do deviations from k * pi/2 for the fourier rotation function? Because if people request 91 degrees of rotation, the actual amount is 90deg (perfect numerical implementation available) + 1 deg (Fourier shear approach) ? Or does the shear approach take this into account?

ehpor:
Only for very specific grids is a perfect 90deg rotation available. I did not implement that optimization. The grid needs to be symmetric around the center. This PR already adds a lot of functionality, so that optimization can be added later.

I'll fix the docstring typo.

PR #181, opened 2023-04-10
Fourier symmetry tests

ehpor:
These new tests check the Fourier symmetries of even and odd real functions, making sure that the resulting FT is either purely real or imaginary respectively. Additionally, these check for rudimentary linearity properties.

Note: compared to the energy conservation and pattern matching checks, I had to remove the input grid shift variable. That is because I don't know an easy way to create an even or odd patterns for testing. Maybe a cosine series with random coefficients? That feels like a lot of work.

### DIFF
--- a/tests/test_fourier.py
+++ b/tests/test_fourier.py
@@ -2,22 +2,25 @@ from hcipy import *
 import numpy as np
 import pytest
 
+def make_all_fourier_transforms(input_grid, q, fov, shift):
+	fft1 = FastFourierTransform(input_grid, q=q, fov=fov, shift=shift, emulate_fftshifts=True)
+	fft2 = FastFourierTransform(input_grid, q=q, fov=fov, shift=shift, emulate_fftshifts=False)
+	mft1 = MatrixFourierTransform(input_grid, fft1.output_grid, precompute_matrices=True, allocate_intermediate=True)
+	mft2 = MatrixFourierTransform(input_grid, fft1.output_grid, precompute_matrices=True, allocate_intermediate=False)
+	mft3 = MatrixFourierTransform(input_grid, fft1.output_grid, precompute_matrices=False, allocate_intermediate=True)
+	mft4 = MatrixFourierTransform(input_grid, fft1.output_grid, precompute_matrices=False, allocate_intermediate=False)
+	nft1 = NaiveFourierTransform(input_grid, fft1.output_grid, precompute_matrices=True)
+	nft2 = NaiveFourierTransform(input_grid, fft1.output_grid, precompute_matrices=False)
+
+	return [fft1, fft2, mft1, mft2, mft3, mft4, nft1, nft2]
+
 def check_energy_conservation(dtype, shift_input, scale, shift_output, q, fov, dims):
 	grid = make_uniform_grid(dims, 1, has_center=True).shifted(shift_input).scaled(scale)
 	f_in = Field(np.random.randn(grid.size), grid).astype(dtype)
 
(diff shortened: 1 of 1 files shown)


syhaffert:
I am a little bit confused. You say that you needed to remove checks for the input shift. But no input_shift check in the file has been removed as far as I can see. Do you mean that the input_shift parameter is not used in the symmetry check?

ehpor:
> I am a little bit confused. You say that you needed to remove checks for the input shift. But no input_shift check in the file has been removed as far as I can see. Do you mean that the input_shift parameter is not used in the symmetry check?

@syhaffert Yes, that's exactly what I mean.

tests/test_fourier.py
```diff
@@ -79,15 +82,62 @@ def test_fourier_energy_conservation_2d(dtype):
...
+
+	# Check symmetry of the input grid.
+	for i in range(input_grid.ndim):
+		assert np.allclose(input_grid.coords[i], -input_grid.coords[i][::-1])
+
+	# Make even and odd input fields.
+	f_in = Field(np.random.randn(input_grid.size), input_grid).astype(dtype)
+
+	f_in_even = (f_in + f_in[::-1]) / 2
+	f_in_odd = (f_in - f_in[::-1]) / 2
```
ivalaginja, line 98: The PR looks good to me, but could I ask you about the logic here? It seems that you add a backwards-ordered function to itself, or subtract from itself, to get an even or odd function, respectively. These do not turn out strictly even or odd, do they? With f(x) = f(-x) (even) and f(x) = -f(-x) ? When I plot them they are even-ish and odd-ish, and yet they obey the fact that evenness remains real only, and oddness gives you a purely imaginary function after the FT, as long as they are both real initially. What am I not understanding here?

ivalaginja: Or, let me rephrase - why do you consider `f_in_even` and `f_in_odd` insufficient for testing, as pointed out by you in the PR description? Once I looked at the actual data across the diagonal of these functions, they are actually indeed even and odd (I was only looking at plots when writing my previous comment, and that was not an accurate enough assessment).

syhaffert: These functions are sufficient for testing. However, these functions are not 'even' and 'odd' anymore when you shift them. So you can't test the shifted input for symmetries. This might be solved by sampling something like a cosine and shifting it by exactly a period?

ehpor: `f_in_even` and `f_in_odd` will only be even and odd for grids with zero being in the middle of the grid, either centered on a pixel (for an odd number of pixels) or in between two pixels (for an even number of pixels). Any other grid, and this way of separating out even and odd parts doesn't work.

Yep, using random coefficients for a few cosines of specific frequencies, evaluated on any grid, might work, as I eluded to. But, while the underlying mathematical function is purely even, it could be a tiny bit odd when evaluated on a sampled, shifted sampled grid, with Gibbs phenomena, sampling things, etc... I'm too lazy to find out if this is the case or not. The current PR should be enough anyway, I'm arguing. This catches most of the problems we might have with symmetries. Additionally, it's something I can point to when someone tells me that HCIPy has an error with symmetry, even though they themselves made the mistake.

FYI, this CI check was planned in early 2018 and I never got around to it.

ivalaginja: Ah I see. Yeah I really appreciate having this test because I sometimes have people object to using hcipy "because you don't know whether it's actually doing the right thing". So this is great to have.

Back in 2019 I was looking into the same thing with sines and cosines but never pushed through with it so there seems to be a commonality here... In any case, I think this PR is good to go.

ivalaginja (approved):
LGTM

PR #187, opened 2023-04-24
New-style Fields

ehpor:
This PR starts the transition to a new implementation of Fields that is more amenable to backend changes later on. We will refer to the current style of Fields as OldStyleField and the new implementation as NewStyleField. The OldStyleFields are based on [subclassing numpy arrays](https://numpy.org/doc/stable/user/basics.subclassing.html). At the time HCIPy was written, subclassing was the only way to associate a Grid with a Field. Subclassing has several disadvantages, most notably that we fully rely on Numpy to "remember" that the array is in fact a Field. This has been a problem early-on, with the `np.angle()` function, which would not properly call `__array_finalize__()`. 

Recent NEPs have made it more and more possible to avoid subclassing, and instead create a completely new class that behaves just like a numpy array. This approach has been successfully used by Dask, CuPy and others, and is now fully mature. The NewStyleField class uses this implementation, with a `Field.data` attribute that contains the values and a `Field.grid` attribute that contains the grid as before.

Since this is a huge transition, that will affect everyone if done improperly, this will be done with feature flags, which will be turned off for now. This allows everyone to test their code, on an opt-in basis, until we deem that the feature is mature enough to enable by default. Even later, the old-style fields and the corresponding feature flag will be removed.

Open questions:
- [x] How to properly test with feature flags? Currently, tests are only run for the default configuration. Tests should most likely be run for both old- and new-style Fields.

### DIFF
--- a/hcipy/config/default_config.yaml
+++ b/hcipy/config/default_config.yaml
@@ -31,3 +31,7 @@ plotting:
 
   # The default colormap for imshow_pupil_phase().
   pupil_phase_colormap: 'RdBu'
+
+core:
+  # Whether to use new-style or old-style fields.
+  use_new_style_fields: false
--- a/hcipy/field/__init__.py
+++ b/hcipy/field/__init__.py
@@ -32,7 +32,8 @@ __all__ = [
     'subsample_field',
     'evaluate_supersampled',
     'make_uniform_vector_field',
-    'make_uniform_vector_field_generator'
+    'make_uniform_vector_field_generator',
+    'is_field',
 ]
 
 from .grid import *
--- a/hcipy/field/field.py
+++ b/hcipy/field/field.py
@@ -211,5 +211,168 @@ def _field_reconstruct(subtype, baseclass, baseshape, basetype):
 
 	return subtype.__new__(subtype, data, grid)
 
-class NewStyleField:
-	pass
+def _unwrap(arg):
+	if isinstance(arg, Field):
+		return arg.data
+
+	if isinstance(arg, list):
+		return [_unwrap(x) for x in arg]
+
+	if isinstance(arg, dict):
+		return {key: _unwrap(val) for key, val in arg.items()}
+
+	if isinstance(arg, tuple):
+		return tuple(_unwrap(x) for x in arg)
+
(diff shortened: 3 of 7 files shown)


ehpor:
@syhaffert Can you mind taking a first quick review pass on this? No rush whatsoever: I'm still working out some kinks on how to test this. But the general structure is there and this is quite an involved transition. All tests pass locally on both new-style and old-style grids.

syhaffert:
I am currently reading up on the new way of sub-classing and interoperability with numpy. When I am done with that I might have more comments. But from what I can see now, I think this implementation is looking pretty good. It's pretty logical.

There are now a couple places where you explicitly cast to an array (np.asarray) is that to make it work with non array type objects?

ehpor:
> There are now a couple places where you explicitly cast to an array (np.asarray) is that to make it work with non array type objects?

Those are actually all cases where it should've been a numpy array in the first place and not a Field. It's just that a Field can end up in there and you wouldn't notice normally if there is no explicit check for that, like is the case currently.

ehpor:
I added unit tests for basic arithmetic and pickling, that is performed for both old- and new-style fields. While this is not a complete end-to-end test, this unit test checks most things that are used in the rest of the codebase.

There are several improvements that can be made still. A major example is the use of Field for after `__array_ufunc__` and `__array_function__`. Recasting into a Field afterwards, in case the resulting object is an array, is insufficient. There are several objects that should have a custom Grid attached, which can be inferred from the arguments to the function. This is clearly a major effort that requires a separate PR. As is, however, the current code mimics the logic that exists for old-style fields. Therefore, I'm fine with merging as is.

Another potential improvements is explicit use of `np.array([])` in `__setstate__()` of `NewStyleField`. For non-Numpy objects, pickling will not work. Right now, I don't know how to solve this, so this will have to wait for a future PR. Additionally, this is not needed as the goal of this PR is purely an underlying infrastructure change rather than adding new functionality that is enabled by the underlying infrastructure changes.

@syhaffert Ready for final review.

syhaffert (approved):
Looks good to me. I think its ready for merging.

PR #192, opened 2023-04-28
Refactor the single mode fiber injection unit

### LINKED ISSUE #191: 'gaussian_mode' is not defined
Looks like the mode basis 'gaussian_mode' is missing. Maybe it was removed? It is called by default in the SingleModeFiber Detector.

ehpor:
Converts the single-mode fiber detector into a single-mode fiber injection optical element.

- [x] Add back the single-mode fiber detector and make it deprecated.
- [x] Add pytest.
- [x] Add new class and function to `__init__.py`.

Fixes #191.


hcipy/optics/fiber.py
```diff
@@ -178,13 +213,84 @@ def read_out(self):
...
+			position = np.zeros(2)
+		else:
+			position = np.array(position)
+
+		self.output_grid = CartesianGrid(RegularCoords([1, 1], [1, 1], position))
+		self.output_grid.weights = 1
+
+		# Compute the fiber mode and normalize.
+		self.mode = mode(self.input_grid.shifted(-position), mode_field_diameter)
+		self.mode /= np.sum(np.abs(self.mode)**2 * self.input_grid.weights)
```
syhaffert: The fiber mode should be normalized by the sqrt of the power.

hcipy/optics/fiber.py
```diff
@@ -178,13 +213,84 @@ def read_out(self):
...
+		----------
+		wavefront : Wavefront
+			The complex amplitude for the single-mode fiber.
+
+		Returns
+		-------
+		Wavefront
+			The outgoing wavefront.
+		'''
+		return Wavefront(wavefront.electric_field * self.mode, wavefront.wavelength, wavefront.input_tokes_vector)
```
syhaffert: typo. tokes -> stokes.

syhaffert (changes requested):
I have two comments that need to be addressed.

ehpor:
@syhaffert This wasn't ready for review. Now it is. Both were fixed.

hcipy/optics/fiber.py
```diff
@@ -3,7 +3,35 @@
...
+	'''
+	def func(grid):
+		if grid.is_('cartesian'):
+			r2 = grid.x**2 + grid.y**2
+		else:
+			r2 = grid.as_('polar').r**2
+
+		# Compute and normalize the Gaussian function.
+		res = np.exp(-r2 / ((0.5 * mode_field_diameter)**2))
+		res /= np.sum(np.abs(res)**2 * grid.weights)
```
syhaffert, line 30: We know theoretically what the normalization factor should be, right? It's 1/sqrt(2*pi * sigma^2) with some correction for the pixel size. Wouldn't that be a better normalization? I ran into issues before where I couldn't rely on numerical normalization due to a finite grid size.

syhaffert (commented):
I only have 1 comment.

hcipy/optics/fiber.py
ehpor: Fixed. There is no correction for pixel size. However, there is a change inside the exponential itself, since we are normalizing for the square, not the Gaussian itself.

BTW, should we then also remove the normalization inside the `SingleModeFiberInjection` element itself? That feels wrong though.

syhaffert: Yeah I think we need to leave the normalization inside that class just in case some weird mode is used there.

hcipy/optics/fiber.py
```diff
@@ -178,13 +214,88 @@ def read_out(self):
...
+		----------
+		wavefront : Wavefront
+			The incoming wavefront.
+
+		Returns
+		-------
+		Wavefront
+			The complex amplitude of the single-mode fiber.
+		'''
+		output = np.dot(wavefront.electric_field * self.input_grid.weights, self.mode)
```
syhaffert, line 258: I think you are missing a complex conjugate here. The coupling into a particular mode is,
```
np.conj(E).dot(E_in)
```
This does not matter too much for Gaussian modes because those are completely real. But you do run into issue if we have arbitrary modes.

hcipy/optics/fiber.py
```diff
@@ -205,15 +316,33 @@ def __init__(self, input_grid, fiber_grid, mode, *args, **kwargs):
...
+		wavefront : Wavefront
+			The incoming wavefront.
+
+		Returns
+		-------
+		Wavefront
+			The complex amplitudes of each fiber.
+		'''
 		res = self.projection_matrix.T.dot(wavefront.electric_field * self.input_grid.weights)
 		return Wavefront(Field(res, self.fiber_grid), wavefront.wavelength)
```
syhaffert, line 332: This also needs a complex conjugate somewhere. Either in the fiber modes or the input field.

syhaffert (changes requested):
I think this is the last issue.

hcipy/optics/fiber.py
```diff
@@ -158,16 +189,21 @@ def backward(self, instance_data, wavefront):
...
-		self.mode /= np.sum(np.abs(self.mode)**2 * self.input_grid.weights)
+		try:
+			self.mode = mode(self.input_grid, mode_field_diameter)
+		except TypeError:
+			self.mode = mode(self.input_grid)
+
+		self.mode /= np.sqrt(np.sum(np.abs(self.mode)**2 * self.input_grid.weights))
 		self.intensity = 0
 
 	def integrate(self, wavefront, dt, weight=1):
```
syhaffert, line 209: Missing complex conjugate here too.

syhaffert (commented):
I found another missing conjugate.

Yeah I think it has been here for a while. That just shows that this part of the code is not used that much. Do we need to add a np.real to the outcome of the dot product? Intensity is by definition real and adding that np.real will make sure it is compliant.

hcipy/optics/fiber.py
syhaffert: its in the integrate function but I couldn't add a comment there because of ????

hcipy/optics/fiber.py
sylacour: The normalisation is important, otherwise, the throughput will look oddly too high. Also, the mode_field_diameter requires a clear definition. It can change in textbooks. Usually, I use the same as Thorlabs: 1/e2 on the power (https://www.thorlabs.com/NewGroupPage9_PF.cfm?ObjectGroup_ID=949). So it would be:

		res = np.exp(-r2 / ((mode_field_diameter)**2))
		res *= np.sqrt(2/np.pi)/mode_field_diameter
		
(from heart, but please double check!)

hcipy/optics/fiber.py
ehpor: Because it's not code that is not changed by this PR. I'll put the conjugate there, this time. But because it's a deprecated class, I don't really want to spend time on fixing stuff that's gonna be removed in the near future.
 People should be using the new functions/classes if they run into bugs.

ehpor: Done.

hcipy/optics/fiber.py
ehpor: I kept the numerical normalization inside the `SingleModeFiberInjection` class.

@sylacour Thanks. I indeed had a wrong definition of the MFD. (I intended to use the 1/e2 one, but I made a mistake in evaluating the Gaussian.) Your from-heart formula was actually correct, except for exchanging MFD with the mode field radius. I also added a test to make sure that the 1/e definition is always checked for this function. The normalization was already tested.

syhaffert (approved):

PR #199, opened 2023-11-03
replace asdf.fits_embed

braingram:
asdf 3.0 removed the deprecated `AsdfInFits` (and `asdf.fits_embed`) used here to read/write fits files (see https://asdf.readthedocs.io/en/3.0.1/asdf/whats_new.html#whats-new-3-0-0 for descriptions of other changes).

This PR replaces uses of `asdf.fits_embed` with equivalent code. This was tested locally by modifying:
- test_grid_io
- test_field_io
- test_mode_basis_io
To save all test files (and not remove them after the test passed). Test files were generated with hcipy main and asdf 2.15.2 (we'll call these 'old' files) and with hcipy using this PR and asdf 3.0.1 ('new' files). hcipy main with asdf 2.15.2 could read both new and old files as could hcipy with this PR and asdf 3.0.1.

If there are static files that aren't captured by these tests it would be helpful to test those files with this PR to verify that the updated code can open all files generated with older versions of hcipy.

### DIFF
--- a/hcipy/util/io.py
+++ b/hcipy/util/io.py
@@ -1,9 +1,9 @@
 import datetime
+import io
 import pickle
 
 import numpy as np
 import asdf
-import asdf.fits_embed
 from astropy import wcs
 from astropy.io import fits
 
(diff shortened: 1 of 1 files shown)


ehpor:
As discussed offline with Brett, the failing tests are likely caused by memmap handling changes between Windows and Linux/Mac. I'll try to fix this since the changes are most likely easy to fix, plus easy to combine with reviewing. If they are difficult to resolve, the ball is back in Brett's court.

ehpor:
@braingram I don't know why codecov reports are not coming through, but it's positive diff coverage, so I'm fine with force merging. Everything is working now as well.

image

ehpor (approved):
LGTM and tested very well.

ehpor:
Force merging since codecov didn't forward its check to Github Checks for some reason. See comment above.

PR #208, opened 2023-11-06
Explicitly use zeroth element of random array.

ehpor:
Conversion of an array with ndim > 0 to a scalar is deprecated in Numpy 1.25.

Fixes https://github.com/ehpor/hcipy/issues/205


syhaffert (approved):
ready for merging

PR #210, opened 2023-11-06
Explicitly use v2 of the imageio API.

ehpor:
This gets rid of deprecation warnings. Note that we can switch to v3 of the imageio at a later point. Let's be conservative with the versioning for now.

This PR also improves code coverage for animations by testing all three ways that frames can be added to an animation, rather than only the default one.

Fixes https://github.com/ehpor/hcipy/issues/209

### DIFF
--- a/hcipy/plotting/animation.py
+++ b/hcipy/plotting/animation.py
@@ -6,7 +6,7 @@ import io
 
 import matplotlib as mpl
 import matplotlib.pyplot as plt
-import imageio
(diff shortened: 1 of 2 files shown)


syhaffert (dismissed):
ready for merging.

ehpor:
@syhaffert Had to fix merge conflicts. Can you re-approve?

syhaffert (approved):
ready for merging.

PR #213, opened 2023-11-06
Add FFT backend support for faster FFTs.

ehpor:
On some platforms (ie. Apple M2), MKL is not available and therefore FFTW is faster than the default Numpy implementation. This PR adds a code path for FFTW in the Fourier transforms.

It adds a config item for the backends that should be attempted and which order they should be attempted. The current ordering of backends is `['mkl', 'fftw', 'numpy']`. This PR adds a math submodule that contains wrapped versions of `fftn()` and `ifftn()` that looks at the current config values and can be overridden.

WIP:
- [x] Not all arguments are available on all backends. `overwrite_x` on `fftn()` is an example, which is only available on MKL and FFTW. This parameter should be ignored on the Numpy backend, but currently throws an error. Note: this is only when the Numpy backend is forced but MKL is still available.
- [x] FFTW backend does not use `overwrite_x` in some cases, even though it could be used and it is faster to use it.
- [x] Use a threshold for automatic multithread vs. single-thread FFTs (when no thread count is given specifically).
- [x] Restrict to PyFFTW <0.13.1 due to multithreading bug in Linux installation.


ehpor:
Benchmark with 1024x1024 FFTs:
```
fftw:
8.48 ms ± 64.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
numpy:
14 ms ± 38.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

ehpor:
New benchmarks of just the FFT (to check overhead of the wrapper). Measurement was done on a 1024x1024 complex128 array.
```
numpy raw
11.8 ms ± 449 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
numpy with hcipy
11.8 ms ± 189 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
fftw raw
6.01 ms ± 194 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
fftw with hcipy
5.86 ms ± 236 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
scipy raw
4.14 ms ± 178 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
scipy with hcipy
4.22 ms ± 244 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

ehpor:
Full benchmark, adapted from Poppy and including Poppy numbers as well:
image

ehpor:
Apparently linux builds of PyFFTW on conda-forge are built without multithreading by accident: error from CI
```
    def _default_threads(threads):
        if threads is None:
            if config.NUM_THREADS <= 0:
                return multiprocessing.cpu_count()
            return config.NUM_THREADS
        else:
            if threads > 1 and _threading_type is None:
>               raise ValueError("threads > 1 requested, but pyFFTW was not built "
E               ValueError: threads > 1 requested, but pyFFTW was not built with multithreaded FFTW.
```

The MacOS Python 3.7 error is different and fails due to somehow importing ctypes.wintypes? Error from CI:
```
___________________ ERROR collecting tests/test_aperture.py ____________________
/usr/local/miniconda/envs/test-environment/lib/python3.7/site-packages/_pytest/runner.py:341: in from_call
    result: Optional[TResult] = func()
/usr/local/miniconda/envs/test-environment/lib/python3.7/site-packages/_pytest/runner.py:372: in <lambda>
    call = CallInfo.from_call(lambda: list(collector.collect()), "collect")
/usr/local/miniconda/envs/test-environment/lib/python3.7/site-packages/_pytest/python.py:531: in collect
    self._inject_setup_module_fixture()
/usr/local/miniconda/envs/test-environment/lib/python3.7/site-packages/_pytest/python.py:545: in _inject_setup_module_fixture
    self.obj, ("setUpModule", "setup_module")
/usr/local/miniconda/envs/test-environment/lib/python3.7/site-packages/_pytest/python.py:310: in obj
    self._obj = obj = self._getobj()
/usr/local/miniconda/envs/test-environment/lib/python3.7/site-packages/_pytest/python.py:528: in _getobj
    return self._importtestmodule()
/usr/local/miniconda/envs/test-environment/lib/python3.7/site-packages/_pytest/python.py:617: in _importtestmodule
    mod = import_path(self.path, mode=importmode, root=self.config.rootpath)
/usr/local/miniconda/envs/test-environment/lib/python3.7/site-packages/_pytest/pathlib.py:567: in import_path
    importlib.import_module(module_name)
/usr/local/miniconda/envs/test-environment/lib/python3.7/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1006: in _gcd_import
    ???
<frozen importlib._bootstrap>:983: in _find_and_load
    ???
<frozen importlib._bootstrap>:967: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:677: in _load_unlocked
    ???
/usr/local/miniconda/envs/test-environment/lib/python3.7/site-packages/_pytest/assertion/rewrite.py:186: in exec_module
    from ..optics import OpticalElement
hcipy/optics/__init__.py:72: in <module>
    from .aberration import *
hcipy/optics/aberration.py:4: in <module>
    from ..util import SpectralNoiseFactoryFFT, inverse_tikhonov
hcipy/util/__init__.py:28: in <module>
    from .spectral_noise import *
hcipy/util/spectral_noise.py:5: in <module>
    from ..fourier import FastFourierTransform, MatrixFourierTransform
hcipy/fourier/__init__.py:16: in <module>
    from .chirp_z_transform import *
hcipy/fourier/chirp_z_transform.py:5: in <module>
    from .._math import fft as _fft_module
hcipy/_math/fft.py:2: in <module>
    from .cpu import get_num_available_cores
hcipy/_math/cpu.py:2: in <module>
    import ctypes.wintypes
/usr/local/miniconda/envs/test-environment/lib/python3.7/ctypes/wintypes.py:20: in <module>
    class VARIANT_BOOL(ctypes._SimpleCData):
E   ValueError: _type_ 'v' not supported
```

ehpor:
No worries, and thanks for letting me know. I'll have someone else review it.

ehpor:
Current test failures are due to MKL not having rfft2 and rfftn implemented. Those should be excluded from the tests.

tests/test_math.py
```diff
@@ -0,0 +1,59 @@
...
+    rng = np.random.default_rng(seed=0)
+
+    print(func, method, dtype_in, dtype_out)
+    if method == 'fftw':
+        pytest.importorskip('pyfftw')
+
+    if method == 'mkl':
+        pytest.importorskip('mkl_fft')
+
+    N = 128
```
raphaelpclt, line 40: This does not test the multi-thread. Is it on purpose?

ehpor: Good catch. This should probably be a for-loop over `N`, with `N` in `[32, 128, 1024]` or something.

raphaelpclt: `[128, 256]` should be enough to keep it minimal?

ehpor: @raphaelpclt Done by [9e2b20c](https://github.com/ehpor/hcipy/pull/213/commits/9e2b20c94c7bdb278a160356a0f3f2cfe9c53dbd).

raphaelpclt (approved):
Looks good on my side.

ehpor:
@raphaelpclt This needs to be reapproved after increasing the tolerance for the single precision tests.

raphaelpclt (approved):
Looks good

ehpor:
Thanks for the review! 🎉

Issue #177, opened 2023-04-03
Bug in LyotCoronagraph if used with focal plane OpticalElement


syhaffert:
There is a bug in the LyotCoronagraph if it is used with an OpticalElement as the focal plane mask. The class tries to automatically guess the coronagraphic focal plane on which the OpticalElement is defined. This is necessary for the Fraunhofer propagation from the pupil to the focal plane. However, when an OpticalElement is used, it tries to get it from the element.input_grid. This does not return the input grid but a function that needs to be evaluated to get the actual input grid. I have attached a minimal working example.

```
from hcipy import *
import numpy as np

if __name__ == "__main__":
	pupil_grid = make_pupil_grid(256, 1.1)
	focal_grid = make_focal_grid(q=5, num_airy=15)
	prop = FraunhoferPropagator(pupil_grid, focal_grid)
	aperture = evaluate_supersampled(make_circular_aperture(1), pupil_grid, 8)

	lyot_fpm = Apodizer(1 - make_circular_aperture(4.0)(focal_grid))
	coronagraph = LyotCoronagraph(pupil_grid, lyot_fpm)

	wf = Wavefront(aperture)
	wf_cor = prop(coronagraph(wf))
```

A proposed work around is to add an optional parameter where you explicitly pass the focal plane mask.

```
def __init__(self, input_grid, focal_plane_mask, lyot_stop=None, focal_length=1, coronagraph_grid=None):
		if coronagraph_grid is None:
			if hasattr(focal_plane_mask, 'input_grid'):
				# Focal plane mask is an optical element.
				coronagraph_grid = focal_plane_mask.input_grid
				self.focal_plane_mask = focal_plane_mask
			else:
				# Focal plane mask is a field.
				coronagraph_grid = focal_plane_mask.grid
				self.focal_plane_mask = Apodizer(focal_plane_mask)

		if lyot_stop is not None and not hasattr(lyot_stop, 'input_grid'):
			lyot_stop = Apodizer(lyot_stop)
		self.lyot_stop = lyot_stop

		self.prop = FraunhoferPropagator(input_grid, coronagraph_grid, focal_length=focal_length)
```

syhaffert:
The second approach has been implemented in #217. That has solved this issue for me. @ehpor do you want to keep this issue open to keep track of the original issue?

PR #196, opened 2023-07-07
Change indentation from tabs to spaces.

ehpor:
This PR changes all indentation from tabs to spaces. This is to conform to PEP8 [(ie. 4 spaces per indentation)](https://peps.python.org/pep-0008/#indentation) and counters a decision that we made at the start of this project.

Outline of the changes:
* First commit changes all tabs to spaces and does nothing else.
* Second commit adds a `.git-blame-ignore-revs` file. This file allows Github to ignore the first commit from all git-blame calls. This avoids cluttering up the git-blame view with just the blame from the first commit.
* Third commit removes the line from the documentation talking about using tabs for indentation.

Note: the Makefile for the documentation still contains tabs, which is required for Makefiles.

PR #217, opened 2023-11-12
Lyot coronagraph fix

syhaffert:
This pull request fixes the bug in #177. I added an explicit parameter to pass the coronagraphic focal grid to the lyot coronagraph. And I have also changed how the grid is determined when an optical element is passed into the function. The function now works with any derivative of the Apodizer class, which should capture 95% of all cases. The other cases will still work if they pass the focal grid explicitely.

### DIFF
--- a/hcipy/coronagraphy/lyot.py
+++ b/hcipy/coronagraphy/lyot.py
@@ -12,6 +12,10 @@ class LyotCoronagraph(OpticalElement):
 	.. [Soummer2007] Soummer et al. 2007, "Fast computation of Lyot-style
 		coronagraph propagation".
 
+	.. notes:: The LyotCoroonagraph tries to automatically get the grid from the focal_plane_mask.
+		However, a bug in AgnosticOpticalElements creates an invalid access which crashes the initialization.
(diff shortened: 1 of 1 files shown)


ehpor:
@syhaffert I've too noticed that tabs still exist in some files. Apparently #196 didn't catch all files (even though it was a replace-all). I'm doing a second PR for that. So please keep the indentation as is. You'll have to a rebase, dropping the whitespace commit (currently the last one).

I'll look at the rest of the PR tomorrow.

syhaffert:
okay done 😄

ehpor:
@syhaffert Nope, not by adding a commit. That still destroys Git Blame. I said, do it with a rebase.

ehpor (changes requested):
If you add a sentence explaining that using this new parameter is part of a workaround, like put that in a `.. notes::` section in the docstring, then I'm fine with it.

Also, what about renaming it to `focal_plane_mask_grid`. Coronagraphic focal grid doesn't feel specific enough to me.

ehpor (approved):

PR #224, opened 2024-03-05
Change indentation from tabs to spaces (again).

ehpor:
Fixing leftovers from https://github.com/ehpor/hcipy/pull/196. I do not know why a "replace-all" didn't catch all tabs, but there were quite a few leftovers.

Similar to before: an entry in the `.git-blame-ignore-revs` file allows Github to ignore the whitespace commit from all git-blame calls. This avoids cluttering up the git-blame view with just the blame from the first commit.


hcipy/coronagraphy/perfect_coronagraph.py
```diff
@@ -4,132 +4,132 @@
...
-		wavelength : scalar
-			The wavelength at which to calculate the transformation matrix.
-
-		Returns
-		-------
-		ndarray
-			The backward transformation_matrix.
-		'''
-		return self.get_transformation_matrix_forward(wavelength)
+    r'''A perfect coronagraph for a certain aperture and order.
```
ivalaginja, line 7: Completely unrelated to this PR, but is this `r` here on purpose? I found one more further below.

hcipy/field/util.py
```diff
@@ -7,540 +7,540 @@
...
-				raise ValueError('You supplied an f_number or (pupil_diameter, focal_length), and forgot to supply a reference wavelength.')
-			else:
-				spatial_resolution = f_number * reference_wavelength
-
-	delta = spatial_resolution / q * np.ones(2)
-	dims = (2 * num_airy * q * np.ones(2)).astype('int')
-	zero = delta * (-dims / 2 + np.mod(dims, 2) * 0.5)
-
-	return CartesianGrid(RegularCoords(delta, dims, zero))
+    r'''Make a grid for a focal plane.
```
ivalaginja, line 118: Is this `r` here in purpose?

ivalaginja (approved):
Looks all fine by me!

hcipy/coronagraphy/perfect_coronagraph.py
ehpor: The `r` is for raw string, which escapes some of the math in this docstring.

hcipy/field/util.py
ehpor: The `r` is for raw string, which escapes some of the math in this docstring.

PR #225, opened 2024-03-06
tilt the position of the actuators.

syhaffert:
This is an extremely minor bug fix for tilted DMs. In the original implementation the pokes themselves were tilted by stretching or compressing the pokes. However, the actuator positions also have to be modified. I added the x, y and z parameters to the actuator position function.

### DIFF
--- a/hcipy/optics/deformable_mirror.py
+++ b/hcipy/optics/deformable_mirror.py
@@ -76,7 +76,7 @@ def make_gaussian_influence_functions(pupil_grid, num_actuators_across_pupil, ac
     ModeBasis
         The influence functions for each of the actuators.
     '''
-    actuator_positions = make_actuator_positions(num_actuators_across_pupil, actuator_spacing)
+    actuator_positions = make_actuator_positions(num_actuators_across_pupil, actuator_spacing, x_tilt, y_tilt, z_tilt)
 
     sigma = actuator_spacing / (np.sqrt((-2 * np.log(crosstalk))))
     cutoff = actuator_spacing / sigma * cutoff


ehpor:
@syhaffert Tests are failing on the actuator positions.

hcipy/optics/deformable_mirror.py
```diff
@@ -76,7 +76,7 @@ def make_gaussian_influence_functions(pupil_grid, num_actuators_across_pupil, ac
     ModeBasis
         The influence functions for each of the actuators.
     '''
-    actuator_positions = make_actuator_positions(num_actuators_across_pupil, actuator_spacing)
+    actuator_positions = make_actuator_positions(num_actuators_across_pupil, actuator_spacing, x_tilt, y_tilt, z_tilt)
```
ehpor, line 79: No. The `actuator_positions` are the positions in the untiled plane. The Gaussian pokes get calculated on the untiled plane first, then transformed by the `transform_poke()` function to a tiled plane. So the current code is correct.

Unless I'm mistaken in this assessment?

syhaffert: @ehpor and me had an in person discussion about this code. The original code does work as intended, it creates the projected influence functions for the DM. What is missing is the projection of the wavefront itself due to the incidence angle. I adjusted the code so that this tilt of the wavefront was added into the DM code. However, after the discussion we decided to create a separate function / class for it. This PR can be closed and left unmerged.

syhaffert:
@ehpor and me had an in person discussion about this code. The original code does work as intended, it creates the projected influence functions for the DM. What is missing is the projection of the wavefront itself due to the incidence angle. I adjusted the tilted DM code in this PR so that this tilt of the wavefront was added. However, after the discussion we decided to create a separate function / class for it. This PR can be closed and left unmerged.

PR #234, opened 2024-10-01
Grid coords should be Fields

ehpor:
Before this PR: `CartesianGrid.x` was a Numpy array, not a Field. This requires the user to cast to a Field after evaluating basic functions like
```
Field(np.sin(grid.x), grid)
```

This PR fixes that problem. The new syntax would be
```
np.sin(grid.x)
```

I'm also fixing type equality tests which were done using `!=` rather than `is not`.


raphaelpclt (approved):
I'm ok with that.

PR #236, opened 2024-10-03
Fix CI regression and update Python version support

ehpor:
Testing on macos-latest now uses Apple M1 chips, so runs on ARM. The required package mkl_fft is not available on ARM. This PR makes mkl_fft an optional dependency for testing.

The distributed FFMpeg of conda now no longer supports the preset argument for some reason. This PR makes the preset argument optional.

This PR also removes support for Python 3.7 & Python 3.8, and adds support for Python 3.12.


raphaelpclt (approved):
Looks fine to me

ivalaginja (approved):
Looks ok to me to

PR #237, opened 2024-10-24
Magellan aperture fix

syhaffert:
The definition of the Magellan aperture's spiders in hcipy were wrong. I have traced down why the error happened and it's due to misinterpretation of an angle definition. This PR fixes the spiders of the Magellan aperture. This aperture definition has been verified with on-sky pupil images.

I have recreated the Magellan fits file for the test_apertures.py script and I include this new file in the PR.


ehpor:
A quick screenshot of the verification with on-sky images would be nice, but not required if you're not able to share a screenshot of that data publicly.

Otherwise, this looks good to me and I'll approve once I hear back from you. @syhaffert

syhaffert:
That is a little bit complicated. The on-sky pupil was compared to a manufacture pupil mask. And I compared the new implementation with the design of that pupil mask. And those now match as shown by the figure below.
image

ehpor (approved):
LGTM

PR #240, opened 2024-11-10
Add loop parameters to GifWriter.

ehpor:
Loops used to be indefinite, but the default parameter for imageio (which uses Pillow under the hood for this) has changed to not loop at all. This PR restores the original behavior.

Also add kwargs to make it possible to modify parameters of imageio without modifying internal hcipy code.

Thanks @julesfowler for reporting this.


ehpor:
Bump @syhaffert @dsdoelman

ehpor:
Bump nr. 2 @dsdoelman @syhaffert

syhaffert (approved):
I think this is ready for merging.

PR #244, opened 2024-11-20
Make unit tests smaller and use tmpdir for file output.

ehpor:
Smaller tests make parallel execution of tests faster. Also, it enables better output when the tests fail.

Using tmpdir for file output also enables running tests in parallel.


syhaffert (approved):
I think this looks good. And the tests still correctly run according to the github pytest results. So I would say ready for merging.

PR #245, opened 2024-11-20
Fix removed `copy_arrays` parameter in asdf 4.0.0.

ehpor:
The `copy_arrays` was deprecated in asdf 3.1.0 (8 months ago) and is now removed (asdf 4.0.0). This PR changes usage for asdf to use the correct parameter depending on asdf version.

Another option is to force asdf>=3.1.0 to be installed, but given the short time since 3.1.0 was released, I'm doing a version check for now.

This PR also removes the Python 3.7 label that should've been removed in #236.

### DIFF
--- a/hcipy/util/io.py
+++ b/hcipy/util/io.py
@@ -1,6 +1,7 @@
 import datetime
 import io
 import pickle
+import packaging
 
 import numpy as np
 import asdf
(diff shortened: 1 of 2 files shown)


ehpor:
The remaining test failures are fixed by #246.

PR #254, opened 2025-01-09
Add backend support for Jax and CuPy

ehpor:
Add support for multiple backends.

Todo:
- [x] Make fallback function structure. @ehpor 
- [ ] Move existing FFT functions to fallback function structure. @ehpor 
- [ ] Move Stokes math (and other locations where numexpr was used) from `wavefront.py` to math submodule. @MatthijsMars 
- [ ] Move MFT math to math submodule. @MatthijsMars 
- [ ] Move FFT grid shift to math submodule. @MatthijsMars 
- [ ] Move Fresnel numexpr code to math submodule. @MatthijsMars 
- [ ] Move DM numexpr code to math submodule. @MatthijsMars 
- [x] Add Poisson and Gamma distributions for Jax backend. @syhaffert 
- [ ] Fix sparse matrices for `ModeBasis`. @ricolandman 
- [ ] Fix sparse matrices for finite difference math. @ricolandman 
- [ ] Interpolators implementation for backends. @ehpor
- [ ] Check all scipy dependencies. @ehpor 
- [ ] Add pytrees for Jax. @ehpor 
- [x] Add automated CI tests with Jax. @ehpor 
- [x] Replace all `numpy` imports with `hcipy.math.numpy` imports.
- [ ] Run CI tests locally with different default backends (using config).
- [ ] Properly define behaviour for dtypes: [link to file](https://github.com/ehpor/hcipy/blob/e90ee1131b79a05e955df5e292960665000fd26f/hcipy/math/backend.py#L31)
- [x] Fix io files looking for `asdf.array` 
- [ ] Check remaining TODOs in files


MatthijsMars:
Switched now to fully using internal numpy everywhere and it passes all tests (except some of the IO tests using asdf) on the numpy backend. Running the tests under 'jax.numpy' backend still results in lots of problems (most of which are trying to do operations with lists/tuples).

MatthijsMars:
Since the JAX random functions are usually applied to `builtins` I added the `jax` random functions to work for builtins [here](https://github.com/ehpor/hcipy/commit/870d47bd57f3cccd1e611c8eadfee5abc1373af5#diff-2bf929689c4413672aaffdba1de524f2df8ac33345970c50b3f9e33a4fef0e28R128). But of course they should also work for builtins on the numpy backend. This is why all  the tests failed now because it tried to import `jax.random` instead of `numpy`. 

@ehpor We should think about how to deal with this dispatching of `builtins` and also custom objects like `field` and `wavefront` to the corresponding backend. (I don't exactly know how this should be done properly).

MatthijsMars:
@ehpor Sebastiaan told me that you might want to replace some of this work by changing to the standard Array API, what were your thoughts on that and can you explain how that would work/interface with the rest of the functionality in HCIpy?

PR #261, opened 2025-05-23
Nan fix

rmnasa:
Hi! I found that the propagators would throw an error if there were any NaNs in an optic sag profile, so added this line to set it to zero outside of the optic diameter. I was simulating a microlens array which is how this came up (the microlens diameter was smaller than the full simulation pupil size so there were nan's in the corners).

### DIFF
--- a/doc/team.rst
+++ b/doc/team.rst
@@ -20,6 +20,7 @@ Direct contributors to code
 * David Kleingeld (`@dskleingeld <https://github.com/dskleingeld>`__), Leiden Observatory
 * Gilles Orban de Xivry (`@GillesOrban <https://github.com/GillesOrban>`__), Université de Liège
 * Brett Graham (`@braingram <https://github.com/braingram>`__), Space Telescope Science Institute
+* Rachel Morgan (`@rmnasa <https://github.com/rmnasa>`__), NASA Ames Research Center
 
 Testing and QA
 --------------
--- a/hcipy/optics/surface_profiles.py
+++ b/hcipy/optics/surface_profiles.py
@@ -60,7 +60,7 @@ def conical_surface_sag(radius_of_curvature, conic_constant=0):
         curvature = 1 / radius_of_curvature
         alpha = (1 + conic_constant) * curvature**2 * r**2
         sag = r**2 / (radius_of_curvature * (1 + np.sqrt(1 - alpha)))
-        sag[np.isnan(sag)] = np.nanmin(sag)  # propagator will throw an error if there are nan's in this definition, this replaces nans with smallest value of sag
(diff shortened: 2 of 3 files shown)


syhaffert:
I think you a have indeed found a small bug. I think the proposed solution is not a catch all solution. For example what if the surface is negative? You can get that by changing the sign of the radius of curvature. Then you suddenly have a weird jump in surface sag.

I think the solution for this is to only evaluate the sag where the alpha parameter has a valid value (alpha <=1). This could be done like this
```
  alpha = (1 + conic_constant) * curvature**2 * r**2
  valid_mask = alpha <= 1
  sag = grid.zeros()
  sag[valid_mask] = r[valid_mask]**2 / (radius_of_curvature * (1 + np.sqrt(1 - alpha[valid_mask])))
```

ehpor:
I agree that an MLA with NaNs should be fixed. Whether that is post-calculation or pre-calculation is even to me. I however disagree with where the fix is implemented right now. The fix is currently in the surface sag generation functions. I see nothing wrong with _this function_ return NaNs, since the surface is indeed undefined at those regions. So, imo these functions are expected to return NaN in those regions.

However, the MLA classes should protect against this, I agree. But there are several ways of doing this. The NaNs could be set to zero, the minimum, the maximum, etc... Which one is a choice. So, this choice should be implemented in the MLA, not the surface sag functions. Say for example by adding a `fill_value` parameter to each of the MLA classes `__init__()` function, which could be either a number or `"max"` or `"min"` which would take the `np.nanmax()` or `np.nanmin()` of the `mla_opd` before filling in any NaNs.

Would that still work for you? @rmnasa

ehpor:
@rmnasa 
> Do you think it makes sense to add this fill_value parameter to SurfaceApodizer and implement the fix there? The parameter could also be set by MLA and passes to the calls to SurfaceApodizer. What do you think?

I didn't think of that, and that sounds even better. Go for it!

ehpor:
@rmnasa Oh, and I like to add all code contributors to the documentation. If you would like to be included there, you should add your name and affiliation in `https://github.com/ehpor/hcipy/blob/master/doc/team.rst` under "Direct contributors to code". No problem if you don't want to.

ehpor:
I don't see that change as a commit. Did you forget to push? (You might need to pull to get my changes before you can push.)

ehpor (approved):

ehpor:
I rethought this. I added the `fill_value` parameter to the surface sag functions instead of to the apodizer classes. The apodizer classes need to be refactored at some point, but for now it's a lot of work to get the fill value implemented in there.

I also added tests for rotational symmetry of the surface sag functions and for the fill value functionality.

ehpor (approved):

PR #269, opened 2025-06-27
Add tests for all grids and coordinates.

ehpor:
This is in preparation for a refactor of grids and coordinates. Furthermore, while writing these tests, we caught _three_ bugs that were undiscovered until now.


syhaffert (approved):
Looks good to me. It's always tricky to spot bugs in less used parts of the code base.

PR #271, opened 2025-07-01
Use matplotlib transforms for `imshow_field()`

ehpor:
## Description

This fixes several issues.

1. Separate code paths for non-linear axes and linear axes. These used two different matplotlib functions which took work to keep the same external interface for `imshow_field()`.
2. Performance issues. The `NonUniformImage` class from Matplotlib is slow. This class was even used for regularly-spaced grids, which could've just used a normal `plt.imshow()`.
3. Canvas sizing for tight bbox. A tight bbox is the default for inline matplotlib figures. A `NonUniformImage` would give the full image extent as the bbox rather than the visible part. This led to a lot of whitespace around plots in inline matplotlib figures for zoomed-in plots. I never worked out how to fix that.
4. Limited modes for interpolation. `NonUniformImage` only supports `'nearest'` and `'bilinear'`. An `AxesImage` however supports any of `'none', 'auto', 'nearest', 'bilinear', 'bicubic', 'spline16', 'spline36', 'hanning', 'hamming', 'hermite', 'kaiser', 'quadric', 'catrom', 'gaussian', 'bessel', 'mitchell', 'sinc', 'lanczos', 'blackman'`. That's a big improvement.
5. Plotting of fields with grids that have either a negative `delta` or descending `x` or `y` values. This silently only showed the middle two pixels and stretched those to the entire axes area. This now works correctly. 

This PR switches to using non-affine Matplotlib transforms for separated grids, and no transforms for regularly-spaced grids. Since regularly-spaced grids are most commonly used in HCIPy user code, this provides a significant speed improvement for `imshow_field()` plots.

## Potential issues for users

Since now an `AxesImage` is returned instead of a `NonUniformImage`, any code using that as a given, needs to be changed. There were a few cases internally for this: `NonUniformImage.set_data(x, y, z)` vs. `AxesImage.set_data(z)` will most likely be the highest impact, often used for animations.

BTW: the docstring of `imshow_field()` incorrectly said that an `AxesImage` was returned. This was incorrect, but will now be correct.

## Screenshots

Before:
image

After:
image

## Todo

- [x] Look through all plots in the built documentation to see if everything looks fine.
- [x] Add test for `imshow_field()` with a separated grid.


ehpor:
Please use a rebase to make the branch up-to-date to master. That avoids having superfluous merge commits in the git history.

I'll rebase now and remove the merge commit that you introduced. I'll probably need reapproval afterwards, depending on how strict Github is.

ehpor:
No reapproval is needed apparently. Github is being lenient today.

Issue #270, opened 2025-07-01
Feature Request: Add support for segment-level Zernike aberrations


GeorgeVassilakis:
I'm working on simulating PSFs for a segmented telescope mirror and need to apply different Zernike aberrations to individual segments. 

As far as I can tell (within documentation and code) HCIPy currently supports:
- Global Zernike aberrations across the entire pupil
- Segment-level piston, tip, and tilt via `SegmentedDeformableMirror`
- But no built-in way to apply arbitrary Zernike modes to individual segments

If there were a method added to `SegmentedDeformableMirror`, it might look something like this:
```
hsm = hcipy.SegmentedDeformableMirror(segments)
hsm.set_segment_zernikes(segment_id=5, zernike_coeffs={4: 30e-9, 5: 20e-9})
```

I'm happy to look into the feasibility or this, but if you might know an easy way to do this, please let me know. Thanks for the help!

PR #272, opened 2025-07-02
Feature/segment zernike modes

GeorgeVassilakis:
Referencing Issue #270, I've added functionality to create hexagonal Zernike modes (Hexike modes) based on the code Susan provided me. To integrate this, I used a combination of regular programming and Claude Code. So, @ehpor, it would be great if you could quickly look over the code before a merge! Thanks so much, let me know if there's anything else I need to address.

An example mirror with 19 segments is shown below, where each segment has a Zernike polynomial corresponding to it's index (SegmentID 0 has 0/Piston, SegmentID 1 has 1/Tip, etc.):
image

A zoomed in image of the piston mode, showing that the polynomials fit snug to the hexagonal segment.
image


GeorgeVassilakis:
@ehpor Thanks for the feedback. This is my first time contributing to an open-source repository (that isn’t my own), so I really appreciate the tips and notes of proper convention (learned what a linter is, lol). Regarding your comments:
1. **Unlicensed Code:** When Susan had emailed on 7/1, she asked:  “I gave him your functions from the cds_pipeline for now. Would you like him to integrate those into `hcipy` and submit a PR or just keep them separate?”. To which you replied: “Depending on the interface, we can include it in HCIPy. I'd create a PR and iterate on Github on the details.” So, I had thought that was explicit permission to use the CDS code and contribute it to HCIPy on GitHub. In the future, I’ll be sure to keenly keep track of what code has a license and what does not.
2. **Code Churn:** Thanks for the heads up. There’s a few things that were unnecessarily changed from previous code that was not directly related to the implementation of the hexike basis, so I’ll undo those changes (although they didn’t change functionality, just some docs (I believe)). 
3. **Hexike class integration:** My implementation indeed was meant to be initially a “tacked-on” addition, mainly because I wasn’t intimately familiar with the codebase, and _really_ didn’t want to break anything or introduce conflicts. I’d be happy to contribute and convert this to a separate class for all the reasons you listed in your comment (Single Responsibility Principle, functionality separation, etc) with the proper improvements (wavelength dependence, caching, etc). Since the hexike polynomials have a decently large computational load, I’ll take a look at your other classes where caches are used as references (since this is all quite new to me). Are there any other design principles you’d like to note or functionalities you’d like to see included?

Thanks for both your patience and help, I appreciate it :)

ehpor:
I don't know if there is really an opportunity to cache things. That's why I said I was not sure about a function vs a class. If you are able to update an existing computed surface deformation when making modifications to the coefficients, or pre-compute the hexike modes for each segment, that would be worth it. But that's a lot of modes that you need to store, so I'm not sure if anyone has the RAM to use that. Also, that's only worth it if people are modifying the hexike coefficients a lot.

So either a `make_hexike_surface_aberration()` or a `HexikeSurfaceAberration` class would be my choice. I don't know what would be better. I'd start with the function and see if there is anything in there that would benefit from being pre-computed if one were to call that function many times with differing hexike coefficients. My guess right now is no, but I haven't thought about it that much.

And don't worry: these things always go through a few cycles of rewriting. The main problem is that it's hard to change the interface later-on, since we want to be backwards compatible for at least two versions. So any mistakes we make now, we have to live with them for a few years.

ehpor:
Since I made the original code for the Hexike function, I can create a PR that adds this. Then you can start a new branch branching off of my PR. I think that's the cleanest option.

PR #274, opened 2025-07-04
Add support for Python 3.13.

ehpor:
Add support for the latest Python version. See https://devguide.python.org/versions/.


ivalaginja (approved):
LGTM

PR #275, opened 2025-07-07
Modal filtered power law error

syhaffert:
I added a high-pass filtered function for power-law errors. I use this type of phase screen very often. The main reason to make it a separate function is that applying a high-mode count modal filter through the original interface can be quite slow. This additional interface uses the FourierFilter with a super-Gaussian high-pass filter.

Changes in this PR
1. A new ```make_high_pass_power_law_error``` function
2. The ```make_general_gaussian_aperture``` function
3. Added tests for both the power law error functions.


hcipy/optics/aberration.py
```diff
@@ -51,6 +52,49 @@ def psd(grid):
...
+        return 1 - make_general_gaussian_aperture(2 * cutoff_frequency, filter_shape_parameter)(fourier_grid)
+
+    ff = FourierFilter(pupil_grid, filter_function, q=2)
+
+    if aperture is None:
+        aperture = make_circular_aperture(diameter)(pupil_grid)
+    aperture_mask = aperture > 0
+
+    phase_screen = make_power_law_error(pupil_grid, 1.0, diameter, exponent=exponent, aperture=None)
+    phase_screen = np.real(ff.forward(phase_screen + 0j))
```
ehpor, line 91: You need the `+ 0j`?

hcipy/optics/aberration.py
```diff
@@ -51,6 +52,49 @@ def psd(grid):
...
+
+    phase_screen_rms = np.std(phase_screen[aperture_mask > 0]) if aperture is not None else np.std(phase_screen)
+    phase_screen *= std / phase_screen_rms
```
ehpor, line 94: I don't really like this style of normalization. You could also calculate the expected RMS for a specific aperture and normalize by that. But, this is the question of: when you use the word RMS, do you mean the RMS on the given aperture or the expected RMS on that aperture. My fear is always that you'll bias your phase screens statistically if you normalize the way this is currently written.

I'm fine with this, but I wanted your opinions on this.

hcipy/aperture/generic.py
```diff
@@ -870,6 +870,32 @@ def func(grid):
...
+        The general Gaussian aperture.
+    '''
+    if center is None:
+        shift = np.zeros(2)
+    else:
+        shift = center * np.ones(2)
+
+    def func(grid):
+        sigma = 0.5 * full_width_half_maximum / (2 * np.log(2))**(1 / (2 * shape_parameter))
+        r = grid.shifted(shift).as_('polar').r / sigma
```
ehpor, line 894: Doesn't this move the wrong way?

hcipy/optics/aberration.py
syhaffert: I understand what you mean. I usually want a specific variance and then I vary that variance myself to explore the range (or I stochastically pick a variance). I don't see the difference between normalizing by std in this way or what we currently do by normalizing on peak to valley. I can also normalize it on peak to valley to make it more similar to the other function.

hcipy/aperture/generic.py
syhaffert: Correct. I switched the sign.

hcipy/optics/aberration.py
ehpor: Adding the `+ 0j` creates an implicit copy of the array. Some FFT backends have specialized functions for a real FFT with a full complex output without an intermediate copy. So by adding 0j, you'd be actively ignoring that potential optimization.

I'm surprised by the warning. That's obviously the case for complex->real casts without taking the real part (which you're doing here). But there shouldn't be a warning for real->complex casts, which are implicit.

hcipy/optics/aberration.py
```diff
@@ -51,6 +52,49 @@ def psd(grid):
...
+
+    if aperture is None:
+        aperture = make_circular_aperture(diameter)(pupil_grid)
+    aperture_mask = aperture > 0
+
+    phase_screen = make_power_law_error(pupil_grid, 1.0, diameter, exponent=exponent, aperture=None)
+    phase_screen = np.real(ff.forward(phase_screen + 0j))
+
+    phase_screen_std = np.std(phase_screen[aperture_mask > 0]) if aperture is not None else np.std(phase_screen)
+    phase_screen *= std / phase_screen_std
```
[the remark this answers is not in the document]
syhaffert: Yes, I hadn't changed the normalization yet because I was waiting for the outcome of our discussion.

ehpor (approved):

PR #276, opened 2025-07-08
Use tuples for grid parameters

ehpor:
This PR changes the `dims`, `zero` and `delta` parameters of `RegularCoords` to be tuples rather than Numpy arrays. These should have been tuples from the start. This now allows all grid dimensions to be computed without using Numpy and use solely Python. This means that future tracing of grids does not need to run Numpy (or any future backend like Jax, CuPy, Dask or something else).

I'm aware that this PR will likely have consequences for user code. This means that we'll leave it for v1.0 rather than the next release.

This PR touches a lot of code throughout the codebase. It also involves being slightly stricter than previously on the input types:
* The input arguments to `Coords` subclasses should now be iterables. Scalars are not allowed anymore.
* The zero is now a mandatory argument.

I've also fixed several cases where `shape` was used instead of `dims`. This will have no influence on users, unless they used grids which had a different number of pixels in x and y. That is a relatively uncommon usage, and mostly reserved for power-users who are likely aware of the effects that that choice has.

I also added documentation for the Chebishev grid function.

### DIFF
--- a/hcipy/atmosphere/infinite_atmospheric_layer.py
+++ b/hcipy/atmosphere/infinite_atmospheric_layer.py
@@ -110,10 +110,10 @@ class InfiniteAtmosphericLayer(AtmosphericLayer):
 
     def _make_stencils(self):
         # Vertical
-        zero = self.input_grid.zero - np.array([0, self.input_grid.delta[1]])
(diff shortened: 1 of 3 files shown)


ivalaginja (approved):
Looks good to me. Loads of zipping, which makes the code quite harder to read, but it's necessary for the changes.

ehpor:
FYI: not merging this before v0.7.0 release. This brings too much change to the table and is reserved for a major release.

PR #277, opened 2025-07-09
Add tests for grid utils and interpolators

ehpor:
This PR adds tests for grid utils and for interpolation. This PR also fixes two bugs:

* A function call to `Field.is_scalar_field` that should have been a property accessor in `make_uniform_vector_field()`.
* The use of the original grid rather than the grid on which the Field generator was evaluated in `make_nearest_interpolator_unstructured()`.


ivalaginja (approved):
LGTM

PR #278, opened 2025-07-09
Add inverted aperture

syhaffert:
This is another small convenience function that I often need to make myself. For example, whenever I make low or high pass filters that are passed to the FourierFilter class. That class requires a function generator so I often need to type out a separate function to do this inversion.

### DIFF
--- a/hcipy/aperture/generic.py
+++ b/hcipy/aperture/generic.py
@@ -478,7 +478,7 @@ def make_inverted_aperture(aperture):
     Field generator
         The inverted aperture.
     '''
-    return lambda grid: Field(1. - aperture(grid), grid)
+    return lambda grid: 1 - aperture(grid)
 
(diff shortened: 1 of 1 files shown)


syhaffert:
Yes, you are right. I did not know that this existed. Should we rename it? If we do, then I can change all instances in the code base for it together with a deprecation warning for the old function handle.

ehpor:
We can rename it. Put the math `f(x) -> 1 - f(x)` in the docstring though. Otherwise, I could think it would be `f(x) -> -f(x)` or `f(x) -> 1 / f(x)`.

ehpor:
Let's rename the old function. Make the old function deprecated (see `dev.py`) and then we can put this in v0.7.0.

ehpor:
You seem to have pushed commits on an outdated branch and then pushed to Github. I'll fix that with a rebase and look at your code afterwards. At a quick glance everything looks good.

PR #286, opened 2025-11-02
add sign fix to orthogonalized

syhaffert:
This is a very small fix that adds a sign correction factor in the ModeBasis orthogonalized property. I saw this in the CDS pipeline and thought it's actually also a good idea to add in general.


syhaffert:
The errors are not created by my code addition. So, I think the code can merge.

ehpor (approved):
I agree. Not sure why I did it in the CDS pipeline but not here.

PR #287, opened 2025-11-11
Fix `overwrite_x` argument for mkl_fft.

ehpor:
mkl_fft changed their interface and no longer uses the `overwrite_x` keyword argument. This PR switches to the new argument.


syhaffert (approved):
approved for merging.

PR #283, opened 2025-08-07
Add Hexike mode basis.

ehpor:
The first of two PRs, splitting up #272. @GeorgeVassilakis FYI.

This PR adds functions to create Hexike modes. This includes:

* `hexike()`: this computes a single Hexike function. This function should be avoided if you want to compute a series of Hexike modes.
* `make_hexike_basis()`: this computes an entire series of Hexike modes. This function should be preferred if many Hexike modes are computed.
* `hexike_ansi()` and `hexike_noll()` for computing a single Hexike mode for a certain ANSI or Noll index.

PR #296, opened 2025-12-12
Add `SegmentedHexikeSurface`

GeorgeVassilakis:
Good evening, Emiel. This PR adds segment-level hexike integration in a way that addresses the design flaws we discussed in PR #272 (which never got merged) and your PR #283, which added the basis function. Below are a summary of the changes!

- I introduced a new standalone optic, `SegmentedHexikeSurface`, rather than modifying `SegmentedDeformableMirror`, addressing your "I'd create a separate class for the aberration" comment in the PR.
- For each segment, the hexike basis is computed on a grid shifted to that segment’s center using `make_hexike_basis()`, then each mode is masked to its segment. The full set is stored as a sparse `ModeBasis` and combined via `linear_combination`, which keeps RAM reasonable and is consistent with how you do the RAM saving in other methods like `hcipy/optics/deformable_mirror.py`.
- The surface is cached the same way as `DeformableMirror.surface`, so repeated propagations with unchanged coefficients don’t redo the basis dot product.
- I also added an optional convenience helper, `make_segment_hexike_surface_from_hex_aperture(...)`, in the spirit of HCIPy’s other `make_*` geometry helpers. It makes a standard hex-segmented aperture into a ready `SegmentedHexikeSurface`, and I’m totally fine dropping or renaming it if you think people should set their own segmented apertures (I set my own in my [HWO-SLAPS](https://github.com/nasa-jpl/hwo-slaps) codebase from my summer internship). If you don't use the helper, you just call the class. Example usage of that helper function is below!

```python
import numpy as np
from hcipy import *

# Define pupil/grid and wavelength.
pupil_grid = make_pupil_grid(256, diameter=7.2)
wavelength = 500e-9

# Build a segmented hexike surface
hexike_surface = make_segment_hexike_surface_from_hex_aperture(
    num_rings=2,
    segment_flat_to_flat=1.4,
    gap_size=0.01,
    pupil_grid=pupil_grid,
    num_modes=10,                 # per segment, Noll ordered
    hexagon_angle=np.pi/2,        # flat-top
)

# Set per-segment coefficients
hexike_surface.set_coefficients_from_dict({
    0: {1: 50e-9/2},   # segment 0, Noll mode 1 (piston)
    3: {4: 20e-9/2},   # segment 3, Noll mode 4, etc.
}, indexing='noll')
```

Happy to tweak naming or interfaces if you’d like anything closer to existing patterns. Let me know what you think!

### DIFF
--- a/hcipy/optics/__init__.py
+++ b/hcipy/optics/__init__.py
@@ -53,6 +53,8 @@ __all__ = [
     'LinearPolarizingBeamSplitter',
     'CircularPolarizingBeamSplitter',
     'SegmentedDeformableMirror',
+    'SegmentedHexikeSurface',
+    'make_segment_hexike_surface_from_hex_aperture',
     'spherical_surface_sag',
     'parabolic_surface_sag',
     'conical_surface_sag',
@@ -87,6 +89,7 @@ from .micro_lens_array import *
 from .periodic_optical_element import *
 from .polarization import *
 from .segmented_mirror import *
+from .segmented_hexike import *
 from .surface_profiles import *
 from .thin_lens import *
 from .tip_tilt_mirror import *
--- a/hcipy/optics/segmented_hexike.py
+++ b/hcipy/optics/segmented_hexike.py
@@ -1,7 +1,7 @@
 import numpy as np
 import scipy.sparse
 
-from ..field import Field, make_hexagonal_grid
+from ..field import make_hexagonal_grid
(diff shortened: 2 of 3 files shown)


ehpor (changes requested):
Tests are good. Interface looks good. Happy to merge after the changes to the docstring.

ehpor (approved):

PR #299, opened 2026-01-14
Array API compliant fields.

ehpor:
# Overview

The [Array API](https://data-apis.org/array-api/latest/index.html) is a relatively new specification that defines a common interface for array libraries in Python. Its primary goal is to allow consumer code to operate seamlessly on arrays originating from multiple computational backends. At present, supported backends include NumPy, CuPy, JAX, PyTorch, Dask, and others. While each backend still has its own idiosyncrasies, the level of convergence has reached a point where meaningful cross-backend interoperability is possible.

Given this maturity, HCIPy is in a good position to adopt the Array API specification. Doing so enables HCIPy to operate natively on arrays from different backends, reduces backend-specific code paths, and aligns the project with the broader Python scientific computing ecosystem.

From a user perspective, the Array API integration is largely transparent. Users typically interact with `Field`s as before, but the underlying array operations are now dispatched through an Array API–compliant namespace. The backend is inferred from the originating array, and all downstream computations consistently use that backend.

## Array API: a minimal introduction

The Array API standardizes *both* array objects and the function namespace used to operate on them. Instead of calling functions from `numpy`, `cupy`, or `jax.numpy` directly, code interacts with a backend-specific namespace that follows a common specification.

For example, given a function `func()` using NumPy explicitly:

```python
import numpy as np

def func(x):
    return np.sin(x)

# Later on in the code...

x_np = np.asarray([1, 2, 3])
print(func(x_np))
```

With the Array API, the same function can be written in a backend-agnostic way:

```python
def func(x):
    xp = x.__array_namespace__()

    return xp.sin(x)
```

Now the function can be used with arrays from any Array API compatible backend. No other code changes are required.

```python
import cupy as cp

x_cp = cp.asarray([1, 2, 3])
print(func(x_cp))
```

The plan is to have HCIPy compatible with this pattern so that HCIPy Fields can be given to `func(x)` as well. Fields can be constructed with any backend as its data. 

```python
import hcipy as hc

arr = np.random.rand(128)
x_hc = hc.Field(arr, grid)

print(func(x_hc))
```

All subsequent operations on `x_hc`—including arithmetic, ufuncs, and propagations—are dispatched through the appropriate Array API namespace. The same code works unchanged if `arr` originates from a different backend (e.g. CuPy or JAX), provided that backend supports the required Array API features.

# High-level implementation details

HCIPy `Field`s conceptually behave like arrays augmented with an associated grid. Based on this observation, this PR redefines `Field` as a first-class Array API–compliant array. Concretely, this introduces:

- A new `Field` class that implements all required attributes and behaviors of an Array API–compatible array.
- A corresponding `FieldNamespace` that implements the full Array API function namespace and serves as the dispatch layer for array operations.

The new `Field` class replaces the existing `NewStyleField` implementation, which was based on an earlier and ultimately abandoned NumPy effort to generalize its API across backends. The current `NewStyleField` class was disabled by default and never saw adoption, so replacement will likely have minimal backward compatibility impact.

One explicit design decision in this PR is the removal of dynamic configuration changes during Field construction. While such configurability can be useful for experimentation, it is not a pattern that is realistically used in production code. More importantly, it introduces a measurable overhead (~150 ns) on *every* `Field` construction. Given that `Field` objects are often created in tight inner loops, this cost is not acceptable.

For testing and comparison purposes, it remains possible to explicitly construct `OldStyleField` or `NewStyleField` instances. Some internal code paths were adjusted to accommodate this separation cleanly. Also note that mixing of `NewStyleField` and `OldStyleField` objects in one operation yields undefined behavior and is not supported.

# Performance considerations

Optical propagation pipelines typically consist of many small array operations chained together. In such workloads, Python-level overhead can easily dominate runtime if not carefully controlled. As a result, this PR pays close attention to minimizing overhead in the Array API integration.

Overhead is evaluated by benchmarking operations on very small arrays, which intentionally amplifies fixed costs such as function dispatch, attribute access, and object construction.

Notable optimizations include:

- Specialized fast paths for unwrapping tuples of `Field`s with two or three elements.
- Caching of `FieldNamespace` instances to avoid repeated construction and lookup costs.
- Automatic generation of function objects for the namespace and array objects.

Preliminary benchmarks indicate that:
- `Field` construction overhead is comparable to (and in some cases lower than) the previous implementation.
- Ufunc dispatch through the Array API introduces only minimal additional overhead relative to direct NumPy calls (~200ns per call).
- End-to-end propagation benchmarks show no statistically significant regression, while enabling multi-backend support.

Detailed benchmark scripts and results will be included in the PR for review.

# TODO (before review)

- [x] Complete implementation of all required Array API attributes on `Field`
- [x] Verify full function coverage against the Array API specification
- [ ] ~Audit backend-specific edge cases (NumPy, CuPy, JAX)~
- [ ] ~Expand performance benchmarks and document methodology~
- [x] Update user-facing documentation to reflect the new Array API model
- [ ] ~Address known correctness issues discovered during early testing~

Edit: deferring crossed-out todo items for future PRs. There are several PRs that depend on this PR, and this PR is correct except for testing which is done in the other PRs. Whenever a future PR discovers an error, it will be fixed there rather than here.


ehpor:
# Simple benchmarks

## Numpy equivalent
```
> a = np.arange(16)
> a + a
289 ns ± 1.65 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
> a * 2
603 ns ± 3.3 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
> np.sin(a)
604 ns ± 3.51 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

## Old style fields
```
> a_field = hcipy.field.OldStyleField(a, None)
353 ns ± 0.996 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
> a_field + a_field
855 ns ± 7.09 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
> a_field * 2
1.1 μs ± 12.9 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
> np.sin(a_field)
1.11 μs ± 4.32 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

## New style fields
```
> a_field = hcipy.field.NewStyleField(a, None)
136 ns ± 1.42 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
> a_field + a_field
495 ns ± 6.19 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
> a_field * 2
837 ns ± 13.3 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
> np.sin(a_field)
834 ns ± 1.66 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
> xp = a_field.__array_namespace__()
330 ns ± 0.451 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
> xp.sin(a_field)
686 ns ± 2.05 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

Note: New style fields equal or beat old style fields in benchmarks in every case, by sometimes 350ns. Furthermore, new style fields only adds 100-250ns in overhead on top of Numpy operations.

syhaffert (commented):
I read through the code and I think at least on a first read through it looks good. Documentation is indeed missing but you are working on it.

ehpor:
Deferring crossed-out todo items for future PRs. There are several PRs that depend on this PR, and this PR is correct except for testing which is done in the other PRs. Whenever a future PR discovers an error, it will be fixed there rather than here.

This PR is now ready for review and merging!

doc/performance.rst
```diff
@@ -0,0 +1,121 @@
...
+
+Now you can create fields using CuPy arrays for GPU acceleration:
+
+.. code-block:: python
+
+   import cupy as cp
+   from hcipy import make_pupil_grid, Field
+
+   # Create a field using CuPy arrays
+   grid = make_pupil_grid(512)
```
MatthijsMars, line 51: But this creates a numpy grid right? Is it converted to the same backend as what you pass to the field? (I know the grid adaptation is a different PR but from this explanation this is unclear to me)

tests/test_field.py
```diff
@@ -254,73 +254,73 @@ def test_grid_supersampled():
...
-        assert np.allclose(a - b, 0)
-        assert np.allclose(a * b, a)
+    a = Field(a_data, grid)
+    b = Field(b_data, grid)
 
-        assert is_field(a + b)
-        assert is_field(a - b)
-        assert is_field(a * b)
-        assert is_field(np.exp(2j * a))
+    xp = a.__array_namespace__()
```
MatthijsMars, line 282: Here you do infer the array namespace but above they are hardcoded to be numpy arrays. We should test this on the different backends

hcipy/field/field.py
```diff
@@ -216,202 +207,956 @@ def _field_reconstruct(subtype, baseclass, baseshape, basetype):
...
+        The corresponding :class:`Grid` on which these values are set. When this is None,
+        this indicates that there is no grid associated with this Field.
     '''
-    def __init__(self, data, grid):
-        self.data = np.asarray(data)
+    __slots__ = ('data', 'grid')
+
+    def __init__(self, data, grid=None):
+        self.data = data.data if isinstance(data, NewStyleField) else data
         self.grid = grid
```
MatthijsMars, line 328: What happens if grid is in a different backend then the field data?

tests/test_field.py
```diff
@@ -254,73 +254,73 @@ def test_grid_supersampled():
...
+    assert is_field(a.repeat(10))
 
-        assert a[0] == 6
-        assert a[1] == 3
+    assert a.size == a.grid.size
+    assert is_field(a.astype('bool'))
+    assert allclose(a.sum(), a_data.sum())
 
-        assert not is_field(np.asarray(a))
+    a[0] = 6
```
MatthijsMars, line 303: These test cases will fail in jax

hcipy/field/field.py
```diff
@@ -216,202 +207,956 @@ def _field_reconstruct(subtype, baseclass, baseshape, basetype):
...
+    'svd',
+)
+
+def _make_namespace(slots):
+    class Namespace:
+        __slots__ = slots
+
+    return Namespace()
+
+# Global cache for FieldNamespace instances.
```
MatthijsMars, line 950: Does all this namespace stuff need to be in the field class file? It seems like a lot of clutter and not very user friendly.

MatthijsMars (commented):
Had a quick overview of the code and most of it seemed reasonable (a few comments). But I really feel like we shouldn't merge this without having at least multi-backend tests on the field class and standard/basic operations on fields.

hcipy/field/field.py
ehpor: Yes, since the namespace and the array object are dependent on each other. This is the classic circular dependency problem, which is solved here by putting them in the same file.

Now, granted, we could split up this file into three: field_base.py, old_style_field.py and new_style_field.py. I don't think that fixes your issue though, since the new_style_field.py file will still be relatively long.

Finally: hcipy uses a lot of files compared to what is common for Python code. This stems from my background of C++, where one file per class (actually two: header + source files) is the norm. So I naturally create more files than necessary. For Python, code should be ordered by module. And since the array class and the array namespace class are strongly dependent, they should be in the same file.

doc/performance.rst
ehpor: Yes, see #305 for the fix for that. The documentation here jumps the gun a little bit until after merging of that PR. Probably juggling four different branches made me confused. You want me to fix this here?

tests/test_field.py
ehpor: Totally agreed but outside of the scope of this PR. This PR solely is about making fields compatible with the array api.

hcipy/field/field.py
ehpor: I'll leave that as undefined behavior. It is a non-goal for the array api standard to allow different array backends to talk to each other (see the last point of the "Non Goals" of https://data-apis.org/array-api/2022.12/purpose_and_scope.html#out-of-scope). I'm not gonna solve that here.

I'm also against adding a specific check or assertion for that case. That takes away performance and you'd need to add those checks _everywhere_.

But this is a great point that I should've documented somewhere.

MatthijsMars: I think this might get very messy then. What will be the way of setting the default backend? Will there be a one-line command we can use to set the default backend in a particular script such that generator functions return objects with the correct array backend? Will users have to convert all these objects to the matching backend themselves? I think if we don't have any checks we should have very clearly described behaviour. People will otherwise mix arrays and get confusing errors.

tests/test_field.py
MatthijsMars: I don't think it's compatible until you can use it on any array API backend (or at least the ones we choose to support) and we need tests in order to ensure this

ehpor: I ran the Array API test suite manually on the current implementation (see results in earlier comments). But I will add those to this PR then. With exclusions on where Numpy itself deviates from the spec.

hcipy/field/field.py
ehpor: That's the point. There will be _no_ default backend. It is the existence of a default that makes those errors possible. If there is a default, it is easy to mix your intended backend with that default. If there is no default, the backend is always explicit and then it's the user who is explicitly giving multiple backends to HCIPy. E.g. explicitly creating a `Grid` with Jax and then creating a `Field` with a Pytorch array and that grid.

FYI: as far as I'm aware right now, there should be only a few points where you need to explicitly define the backend. Most notably for creating grids, see #305. The rest of the cases, the backend can be inferred from the input arguments of classes/functions.

tests/test_field.py
ehpor: I am aware. This is one of the idiosyncrasies that I talked about in the posts above. See the spec (https://data-apis.org/array-api/latest/design_topics/copies_views_and_mutation.html) for a discussion on this specific one.  The spec specifically allows in-place assignment, so it's Jax who is deviating here, not us. Fixing this would require a special code-path for Jax, which is outside of the scope of this PR.

ehpor:
@MatthijsMars Added the array API test suite. I'm not gonna make it mandatory for a PR to pass yet, since this feature is experimental and hidden behind a feature flag. But at least testing is done during CI and counts towards test coverage.

tests/test_field.py
ehpor: Added.

MatthijsMars (commented):
Great job on adding the testing, I think this strengthens this PR a lot. So just for my understanding and documenting reasons. 

This PR makes Field a proper array_api compliant class. And this is tested using the array-api-strict implementation that implements a minimal array_api backend that can be used for testing purposes. In CI we now test whether the field class is compliant with the Array API formalism and it passes most test and test coverage is include in the codecov. Some tests are excluded as specified in `.github/workflows/array-api-tests-skips.txt`.

I think this is where I misunderstood the intent of this PR slightly. This PR does not implement the Field class now works in the specific additional backends that we want to support (jax, cupy, torch). Since these use (some parts of) the Array API, most operations from these backends should now work on Fields. Though parts where backends deviate from the Array API (such as jax Arrays being immutable) will fail some of our tests and implementations. 

I have one remaining question: Currently all the Field tests (like in test_field) test in the numpy backend. Should we additionally test whether non-standard operations that we implement for fields are specifically working with array api (like testing with array-api-strict) as well?

Further PRs will then implement specific compliance with different backends. 

Does this sum it up correctly? Then I think this PR is almost done

ehpor:
@MatthijsMars Mostly correct.

This PR implements `NewStyleField`, which is a subclass of the `FieldBase` class. It's a specific implementation, which is _not_ used by default. When creating a `Field`, the implementation looks at a feature flag, contained in the config and settable by the user, and chooses between creating a `NewStyleField` or an `OldStyleField`.

The `OldStyleField` (originally just called `Field`) is based on inheriting from `numpy.ndarray`, see [here](https://numpy.org/doc/2.4/user/basics.subclassing.html). By implementing specific functions (`__array_wrap__` and others), Numpy can use these to return `Field` objects from calls to standard Numpy functions (e.g. `numpy.dot()`). This was the only way to implement metadata at the time HCIPy was created.

There are however a lot of downsides to this approach, and so Numpy created a dispatch mechanism (`__array_function__` and others, see [NEP18](https://numpy.org/neps/nep-0018-array-function-protocol.html)). The intent was to eventually deprecate the subclassing method that was used at the time, so I started working on a version that used the dispatch mechanism instead, under a deactivated feature flag to not inconvenience current users and still allow power users to test it. With the slow upgrade cycle of astronomical software, HCIPy users were not happy to upgrade to newer Numpy versions too soon, plus deprecation of subclassing has not been announced, so I didn't put a lot of time in to finish that feature. Additionally, the dispatch mechanism still didn't fix some issues that the subclassing method had, so I wasn't too motivated either.

Then the Array API specification came about, which garnered much wider support as a unifying approach for major array packages. It is now clear to me that this spec is the future of scientific Python for the next ten years. This PR changes the "hidden" `NewStyleField` to an implementation of `Field` compatible the Array API spec. That is its sole purpose.

As of this PR, the vast majority of array operations should work on `NewStyleField` objects using any of the backends you mentioned. However, each backend has edge cases wrt the Array API spec, think ten or so per backend. You will only find out what these are when you start running tests with each backend. How to resolve each of these edge cases should be decided on an individual basis. That is out of scope for this PR, hence why I opposed testing with these backends in this PR in the first place. Which subset of edge cases we decide to fix/wrap inside HCIPy depends on usage patterns inside and outside of HCIPy, which we won't discover until we try to run code with `NewStyleFields`. Furthermore, some edge cases are actively being fixed by the backends themselves (including Numpy).

Let's take your example: `__setitem__` is not supported on Jax arrays, since Jax arrays are inherently immutable. Therefore, imo, people using Jax arrays should not expected `a[1] = 4` to magically work, when they "wrap" their Jax array into a `NewStyleField`. However, some code inside HCIPy still expects this operation to work, and so we need to have an alternative code path for those cases.

This PR is not the be-all and end-all of supporting arbitrary backends; that would be a 10-25k line PR, which is unreviewable and allows no splitting of work during development. Imo any PR >1k lines (such as this one) is already too large and takes too long to review properly. Further PRs are/will be made for:
* making `Grids` compatible with arbitrary Array API backends (#305)
* transforming each submodule (math, FFT, propagations, optical elements, apertures, etc.) in HCIPy to only use Array API functions (e.g. #302 but there will be about ten more)
* test infrastructure: using `array_api_strict` and/or other Array API backends, within time constraints on CI. (tbd when most core functionality works, but the test fixtures in #301 is a first step)
* performance PRs, if certain operations are too slow on certain backends (particularly GPUs).
* probably other minor ones that I'm not thinking of right now.

ehpor:
CI failure of Ubuntu Python 3.10 is caused by Numpy being downgraded from Numpy 2.2 to Numpy 1.24, caused by #301 and the newly installed Pytorch library, I think. I'll investigate.

ehpor:
I've had to drop support for Python 3.10, due to the inability to install both Jax and Pytorch at the same time while retaining Numpy 2.0 or higher (which is required for Array API support. While it probably is possible to change the testing infrastructure to use separate conda environments for testing the different parts of the `RandomGenerator` functionality, but that is a lot of work just to support a Python version that we would drop support for in five months time.

I also decided not to enforce Numpy>=2.0. There might be a valid reason to use that still alongside HCIPy, and it is only needed for `NewStyleFields` which are not enabled by default and should be considered experimental for now.

ehpor:
@MatthijsMars Ready for re-review.

I'll remove the repo-required Python 3.10 status checks after you approve this PR.

MatthijsMars (approved):
Looks good to me

PR #301, opened 2026-02-05
Add RandomGenerator class that unifies randomness across backends.

ehpor:
Related to #299 but independent.

This PR adds a `RandomGenerator` class. The `RandomGenerator` class provides a unified and consistent API for generating random numbers across various array API namespaces such as NumPy, CuPy, PyTorch, and JAX. Its primary purpose is to abstract away the specific implementations and nuances of random number generation in these different libraries, allowing users to write generic code that works regardless of the underlying array backend. It initializes an appropriate random number generator based on the provided namespace and an optional seed.

The class exposes common methods for sampling from statistical distributions, including `normal()` (Gaussian), `poisson()`, and `gamma()`. It internally manages the state of the random number generator, handling differences such as explicit PRNG key splitting for JAX's functional paradigm or state management for PyTorch. Additionally, it includes a `copy()` method to create an independent duplicate of the current random state.

I'm also sneaking in a Ruff config change to ignore python 3.8 warnings (which is a Python version that is no longer supported). It was annoying me, sorry.

Also, since most backends are not tested by CI, the coverage will go down despite my best efforts to add comprehensive tests. Tests will be performed if the library is installed.

### DIFF
--- a/hcipy/_math/random.py
+++ b/hcipy/_math/random.py
@@ -1,7 +1,7 @@
 from array_api_compat import is_numpy_namespace, is_torch_namespace, is_jax_namespace, is_cupy_namespace
 import copy
 
-class RandomState:
+class RandomGenerator:
     '''A class that provides a consistent API for random number generation across different array API namespaces.
 
     Parameters
@@ -28,20 +28,21 @@ class RandomState:
                 self._rng.manual_seed(seed)
         elif is_jax_namespace(xp):
             from jax import random
+
             self._jax_random = random
             self._rng = random.PRNGKey(seed or 0)
         else:
             raise ValueError(f"Unsupported namespace: {xp}")
 
     def copy(self):
-        '''Return a new RandomState object that is an independent copy of the current state.
+        '''Return a new RandomGenerator object that is an independent copy of the current state.
 
         Returns
         -------
-        RandomState
-            A new RandomState object with the same internal state as the current object.
+        RandomGenerator
+            A new RandomGenerator object with the same internal state as the current object.
         '''
-        new_rng = RandomState.__new__(RandomState)
+        new_rng = RandomGenerator.__new__(RandomGenerator)
         new_rng.xp = self.xp
         new_rng._jax_random = self._jax_random
 
--- a/hcipy/_math/random.py
+++ b/hcipy/_math/random.py
@@ -1,0 +1,141 @@
+from array_api_compat import is_numpy_namespace, is_torch_namespace, is_jax_namespace, is_cupy_namespace
+import copy
+
+class RandomState:
+    '''A class that provides a consistent API for random number generation across different array API namespaces.
+
+    Parameters
+    ----------
+    xp : object
+        The array API namespace (e.g., numpy, torch, jax) to use for random number generation.
+    seed : int, optional
+        The seed for the random number generator. If ``None``, a random seed will be used, by default None.
+
+    Raises
+    ------
+    ValueError
+        If an unsupported array API namespace is provided.
+    '''
+    def __init__(self, xp, seed=None):
+        self.xp = xp
+        self._jax_random = None
+
+        if is_numpy_namespace(xp) or is_cupy_namespace(xp):
+            self._rng = xp.random.default_rng(seed)
+        elif is_torch_namespace(xp):
+            self._rng = xp.Generator()
+            if seed is not None:
+                self._rng.manual_seed(seed)
+        elif is_jax_namespace(xp):
+            from jax import random
+            self._jax_random = random
+            self._rng = random.PRNGKey(seed or 0)
+        else:
+            raise ValueError(f"Unsupported namespace: {xp}")
+
+    def copy(self):
+        '''Return a new RandomState object that is an independent copy of the current state.
+
+        Returns
+        -------
+        RandomState
+            A new RandomState object with the same internal state as the current object.
+        '''
+        new_rng = RandomState.__new__(RandomState)
+        new_rng.xp = self.xp
+        new_rng._jax_random = self._jax_random
+
+        if is_numpy_namespace(self.xp) or is_cupy_namespace(self.xp):
+            new_rng._rng = copy.deepcopy(self._rng)
+        elif is_torch_namespace(self.xp):
+            new_rng._rng = self.xp.Generator()
+            new_rng._rng.set_state(self.xp.get_rng_state())
+        elif is_jax_namespace(self.xp):
+            new_rng._rng = self._rng  # Immutable, safe to share reference
+
+        return new_rng
+
+    def normal(self, mean=0.0, std=1.0, *, size=None):
+        """Generate random samples from a normal (Gaussian) distribution.
+
+        Parameters
+        ----------
+        mean : float, optional
+            Mean ("center") of the distribution, by default 0.0.
+        std : float, optional
+            Standard deviation (spread or "width") of the distribution, by default 1.0.
+        size : int or tuple of ints, optional
+            Output shape. If ``None``, a single scalar value is returned, by default None.
+
+        Returns
+        -------
+        array
+            An array of specified shape filled with random samples from the
+            normal distribution.
+        """
+        if size is None:
+            size = ()
+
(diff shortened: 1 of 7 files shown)


tests/test_math.py
```diff
@@ -55,3 +56,78 @@ def test_fft_acceleration(func, method, dtype_in, dtype_out):
 
         assert np.allclose(y_numpy, y_method, rtol=rtol, atol=rtol)
         assert y_method.dtype == dtype_out
+
+
+def test_random_generator_distribution(xp):
+    # Get samples from the Standard Normal distribution.
+    rng = RandomGenerator(xp, seed=42)
+    samples = rng.normal(size=(10000,))
```
MatthijsMars, line 64: Should probably test this on non-default mean and standard deviation to show that also mean and standard deviation are propagated properly

tests/test_math.py
```diff
@@ -55,3 +56,78 @@ def test_fft_acceleration(func, method, dtype_in, dtype_out):
...
+def test_random_generator_gamma(xp):
+    # Get samples from the Gamma distribution.
+    rng = RandomGenerator(xp, seed=42)
+    samples = rng.gamma(scale=2.0, shape_param=2.0, size=(10000,))
+
+    # Basic statistical tests (mean should be shape * scale = 2 * 2 = 4)
+    assert np.isclose(float(xp.mean(samples)), 4.0, atol=0.5)
+
+
+def test_random_generator_reproducible(xp):
```
MatthijsMars, line 89: I think the bottom few tests could be parametrised for the different functions to make sure gamma and poisson also follow this behaviour.

MatthijsMars (commented):
And yes I think we should most definitely add jax and pytorch to the test environment. Easiest would probably be to just have two environments, one for windows and one for unix. Or we could probably add a step in the CI that checks if it is not windows and then install pytorch and jax. Then all the code (or at least what has currently been adapted to the multibackend) should be tested in Jax, cupy and pytorch if installed in the environment.

tests/test_math.py
ehpor: Great point. I'll change that. Same for the other two distributions.

hcipy/_math/random.py
```diff
@@ -0,0 +1,215 @@
...
+        array
+            An array of specified shape filled with random samples from the
+            normal distribution.
+        """
+        if size is None:
+            size = (1,)
+        elif not isinstance(size, tuple):
+            size = (size,)
+
+        if is_numpy_namespace(self.xp) or is_cupy_namespace(self.xp):
```
MatthijsMars: Is it necessary to do this if-else every time we generate a new random number? The namespace is inferred only at __init__ so we can also do all the if-else in the init or some other function that constructs the corresponding function links or lambda functions if necessary. 

(same for all other distributions)

MatthijsMars (commented):
I checked the CI and it now did install the optional dependencies (like torch and jax) and the conftest showed that cupy wasn't installed on one of them so that also works. Are the random generator tests in the CI now also actually running for each of the different backends? If not, should we include this? I think turning on multiple backend testing for the whole codebase is right now not a great option but we should automatically test the parts of the code that are compatible and implemented now already.

hcipy/_math/random.py
ehpor: I don't think that would speed things up much. These checks are ~50ns each. But I'll do some benchmarks of the whole implementation:

```python
rng1 = make_random_generator(np)  # new
rng2 = RandomGenerator(np)        # old

print('normal')
%timeit rng1.normal(size=(1,))
%timeit rng2.normal(size=(1,))
%timeit rng1.normal(size=(100,))
%timeit rng2.normal(size=(100,))
%timeit rng1.normal(size=(100,100))
%timeit rng2.normal(size=(100,100))
print('poisson')
%timeit rng1.poisson(size=(1,))
%timeit rng2.poisson(size=(1,))
%timeit rng1.poisson(size=(100,))
%timeit rng2.poisson(size=(100,))
%timeit rng1.poisson(size=(100,100))
%timeit rng2.poisson(size=(100,100))
print('gamma')
%timeit rng1.gamma(size=(1,))
%timeit rng2.gamma(size=(1,))
%timeit rng1.gamma(size=(100,))
%timeit rng2.gamma(size=(100,))
%timeit rng1.gamma(size=(100,100))
%timeit rng2.gamma(size=(100,100))
```
yields
```
normal
621 ns ± 1.83 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
662 ns ± 2.17 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
1.09 μs ± 3.91 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
1.11 μs ± 7.05 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
47.8 μs ± 877 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
47.4 μs ± 76.5 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
poisson
536 ns ± 1.65 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
563 ns ± 2.72 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
2.99 μs ± 25.3 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
3.02 μs ± 11.8 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
249 μs ± 3.71 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
248 μs ± 1.33 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
gamma
631 ns ± 3.84 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
671 ns ± 9.8 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
1.11 μs ± 7.71 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
1.15 μs ± 6.5 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
49.2 μs ± 203 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
49.4 μs ± 54.4 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

So 30-50ns reduction per function call, at the cost of 60 extra lines of code.

ehpor:
@MatthijsMars yes, the tests for the RandomGenerator class are running for each backend that is installed. This is done using a pytest fixture `xp`. Backends that are not installed are skipped with a message indicating that they are not installed. You can check this in the code coverage (see the first comment, or [this direct link](https://app.codecov.io/gh/ehpor/hcipy/pull/301?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Emiel+Por)).

hcipy/_math/random.py
```diff
@@ -0,0 +1,274 @@
...
+
+        self._rng = self._xp.random.default_rng(seed)
+
+    def copy(self):
+        res = super().copy()
+        res._rng = copy.deepcopy(self._rng)
+
+        return res
+
+    def normal(self, mean=0.0, std=1.0, *, size=None):
```
MatthijsMars, line 194: As mentioned, I think there should be docstrings here and for all other functions.

MatthijsMars (commented):
@ehpor Implementation and testing now look good to me. Only thing I would add is copying the copy, normal, gamma, etc. function docstrings to all the dedicated backend classes so each derived class has the proper documentation. Then random generator objects for a specific backend will have proper documentation

ehpor:
@MatthijsMars docstrings for functions inherit from the parent class (since Python 3.5).

MatthijsMars:
@ehpor My bad. Then it looks good to me. I'll approve it for merging

MatthijsMars (dismissed):

ehpor:
@MatthijsMars I accidentally did a merge update rather than a rebase update (stupid phone buttons). Fixed now. Can you reapprove?

MatthijsMars (approved):

PR #302, opened 2026-02-07
Make FFTs Array API compatible

ehpor:
> [!WARNING]  
> Dependent on #299 and #301. Until those are merged, this PR will include additional commits / will fail CI.

This PR makes the _math submodule Array API compatible. This consists of deferring FFT function calls on non-Numpy operations to their array namespace implementation, and using the accelerated HCIPy FFT functions in the FieldNamespace.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -14,6 +14,7 @@ The banner above shows a number of simulations done with HCIPy. From left to rig
    installation
    getting_started/index
    tutorials/index
+   performance
    citing
    team
    changelog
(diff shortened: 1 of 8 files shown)


MatthijsMars (approved):
Looks good to me. Fields are still array API compliant. Original hcipy functionality remains intact for fields that use numpy backend and passes tests.

Next steps for FFTs will be to make it functional for other backends and add testing for this right? (though this will only happen once the full array API update is done I reckon)

PR #305, opened 2026-02-20
Make grids Array API compatible.

ehpor:
> [!WARNING]  
> Dependent on #299 and #301. Until that PR is merged, this PR will include additional commits and will fail CI.

This PR makes Grids compatible with Array API.

### DIFF
--- a/hcipy/_math/backends.py
+++ b/hcipy/_math/backends.py
@@ -3,7 +3,7 @@ import array_api_compat
 from ..config import Configuration
 
 
-def _infer_xp(*arrays):
+def infer_xp(*arrays):
     """Infer xp from input arrays.
 
     Parameters
@@ -23,7 +23,7 @@ def _infer_xp(*arrays):
     """
     # Filter out None values
     arrays = tuple(a for a in arrays if a is not None)
-    
+
     if len(arrays) > 0:
         try:
             return array_api_compat.array_namespace(*arrays)
@@ -82,3 +82,24 @@ def to_numpy(arr):
         return np.asarray(arr)
     except Exception as e:
         raise ValueError(f"Cannot convert {type(arr).__name__} to numpy: {e}")
+
+def is_scalar(element):
+    '''Check whether an element is a scalar.
+
+    Parameters
+    ----------
+    element : any
+        The element to check.
+
+    Returns
+    -------
+    bool
+        Whether the element is a scalar.
+    '''
+    ndim = getattr(element, 'ndim', None)
+    if ndim is None:
+        # Plain Python scalar, no array structure
+        return True
+
+    shape = getattr(element, 'shape', ())
+    return all(n == 1 for n in shape)
--- a/hcipy/_math/backends.py
+++ b/hcipy/_math/backends.py
@@ -1,6 +1,7 @@
 import numpy as np
 import array_api_compat
 from ..config import Configuration
+import sys
 
 def infer_xp(*arrays):
     """Infer xp from input arrays.
@@ -120,7 +121,73 @@ def all_close(a, b, *, rtol=1e-5, atol=1e-8):
     atol : float, optional
         _description_, by default 1e-8
     '''
-    xp = array_api_compat.array_namespace(a, b)
+    xp = array_namespace(a, b)
 
     close = abs(a - b) <= atol + rtol * abs(b)
     return xp.all(close)
+
+namespace_caches = {}
+NUMPY_NAMESPACE = 'numpy'
+
+def array_namespace(*xs, api_version=None):
+    '''Return the namespace of a set of Arrays.
+
+    Parameters
+    ----------
+    *xs : tuple of Arrays
+        The array objects of which to find the namespace.
+    api_version : str, optional
+        The Array API spec version. If this is None (default),
+        the choice is left to the Array object.
+
+    Returns
+    -------
+    module
+        The Array namespace.
+    '''
+    from ..field import NewStyleField, make_field_namespace
+
+    try:
+        cache = namespace_caches[api_version]
+    except KeyError:
+        namespace_caches[api_version] = {}
+        cache = namespace_caches[api_version]
+
+    for x in xs:
+        if isinstance(x, NewStyleField):
+            return make_field_namespace(array_namespace(x.data, api_version=api_version))
+
+        try:
+            ns = cache[type(x)]
+
+            if ns == NUMPY_NAMESPACE:
+                # Could be either a Numpy array or
+                # a JAX zero gradient array.
+                if 'jax' in sys.modules:
+                    import jax
+
+                    if x.dtype == jax.float0:
+                        import jax.numpy as jnp
+                        return jnp
+                else:
+                    return np
+
+            return ns
+        except KeyError:
+            continue
+
+    for x in xs:
+        try:
+            ns = array_api_compat.array_namespace(x, api_version=api_version)
+
+            # Numpy is a special case, see above.
+            if ns == np:
+                cache[type(x)] = NUMPY_NAMESPACE
+            else:
+                cache[type(x)] = ns
+
+            return array_namespace(x, api_version=api_version)
+        except KeyError:
+            cache[type(x)] = None
+
+    return array_api_compat.array_namespace(*xs, api_version=api_version)
--- a/hcipy/_math/backends.py
+++ b/hcipy/_math/backends.py
@@ -181,7 +181,7 @@ def array_namespace(*xs, api_version=None):
             ns = array_api_compat.array_namespace(x, api_version=api_version)
 
             # Numpy is a special case, see above.
-            if ns == np:
+            if x.__class__.__module__.startswith('numpy'):
                 cache[type(x)] = NUMPY_NAMESPACE
             else:
                 cache[type(x)] = ns
--- a/hcipy/_math/backends.py
+++ b/hcipy/_math/backends.py
@@ -191,3 +191,23 @@ def array_namespace(*xs, api_version=None):
             cache[type(x)] = None
 
     return array_api_compat.array_namespace(*xs, api_version=api_version)
+
+def default_dtype(xp, kind, device=None):
+    '''Find the default dtype of a certain kind.
+
+    Parameters
+    ----------
+    xp : module
+        The Array API backend.
+    kind : {"real floating", "complex floating", "integral", "indexing"}
+        The kind of dtype.
+    device : Device or None
+        The device on the Array API backend. If not given, the default device
+        of the backend will be used.
+
+    Returns
+    -------
+    dtype
+        The default dtype.
+    '''
+    return xp.__array_namespace_info__().default_dtypes(device=device)[kind]
--- a/hcipy/_math/backends.py
+++ b/hcipy/_math/backends.py
@@ -2,7 +2,6 @@ import numpy as np
 import array_api_compat
 from ..config import Configuration
 
-
 def infer_xp(*arrays):
     """Infer xp from input arrays.
 
@@ -34,7 +33,6 @@ def infer_xp(*arrays):
     if Configuration().core.use_new_style_fields:
         raise ValueError("xp must be specified when arrays don't provide it")
     return np
-
 
 def to_numpy(arr):
     """Convert any array to numpy.
@@ -103,3 +101,26 @@ def is_scalar(element):
 
     shape = getattr(element, 'shape', ())
     return all(n == 1 for n in shape)
+
+def all_close(a, b, *, rtol=1e-5, atol=1e-8):
+    '''Whether the two arrays are elementwise equal within tolerance.
+
+    This is defined as `|a - b| <= atol + rtol * |b|`.
+
+    At least `a` or `b` needs to be an Array.
+
+    Parameters
+    ----------
+    a : Array or scalar
+        The first array or scalar.
+    b : Array or scalar
+        The second array or scalar.
+    rtol : float, optional
+        _description_, by default 1e-5
+    atol : float, optional
+        _description_, by default 1e-8
+    '''
+    xp = array_api_compat.array_namespace(a, b)
+
+    close = abs(a - b) <= atol + rtol * abs(b)
+    return xp.all(close)
--- a/hcipy/_math/backends.py
+++ b/hcipy/_math/backends.py
@@ -169,8 +169,7 @@ def array_namespace(*xs, api_version=None):
                     if x.dtype == jax.float0:
                         import jax.numpy as jnp
                         return jnp
-                else:
-                    return np
+                return np
 
             return ns
         except KeyError:
--- a/hcipy/_math/backends.py
+++ b/hcipy/_math/backends.py
@@ -2,6 +2,7 @@ import numpy as np
 import array_api_compat
 from ..config import Configuration
 import sys
+from typing import Any
 
 def infer_xp(*arrays):
     """Infer xp from input arrays.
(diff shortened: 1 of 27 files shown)


ehpor:
Manual test result of grids when `use_new_style_fields` is True in the config:

```
pytest tests/test_grid.py 
============================================================================= test session starts ==============================================================================
platform darwin -- Python 3.13.12, pytest-9.0.3, pluggy-1.6.0
rootdir: /Users/epor/Github/hcipy
configfile: pyproject.toml
plugins: cov-7.1.0, xdist-3.8.0
collected 108 items                                                                                                                                                            

tests/test_grid.py .ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss.                                          [100%]

=============================================================================== warnings summary ===============================================================================
tests/test_grid.py::test_polar_grid_transformations[numpy]
tests/test_grid.py::test_polar_grid_transformations[torch]
  /Users/epor/Github/hcipy/hcipy/field/polar_grid.py:40: UserWarning: No automatic weights could be calculated for this grid.
    self.weights *= np.abs(scale)**(self.ndim)

tests/test_grid.py::test_evaluate_supersampled[torch]
tests/test_grid.py::test_evaluate_supersampled[torch]
  /Users/epor/Github/hcipy/hcipy/field/util.py:658: DeprecationWarning: __array_wrap__ must accept context and return_scalar arguments (positionally) in the future. (Deprecated NumPy 2.0)
    dithered_separated_coords = [c + d * delta for c, d, delta in zip(separated_coords, dither, deltas)]

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================================================================== 54 passed, 54 skipped, 4 warnings in 0.54s ==================================================================
```

(with Numpy and Torch in the environment)

hcipy/field/util.py
```diff
@@ -52,44 +55,71 @@ def _find_common_size(*args):
...
     ----------
     dims : int or tuple of ints
         The number of points in each dimension. If this is a scalar, it will
         be multiplexed over all dimensions.
-    extent : scalar or tuple of floats
-        The total extent of the grid in each dimension.
-    center : scalar or tuple of floats
-        The center point. The grid will by symmetric around this point.
+    extent : scalar or array_like
+        The total extent of the grid in each dimension. Can be an xp array for differentiability.
```
MatthijsMars: Do we need to mention this at all of these functions? Might be confusing for people who don't want to use or don't understand differentiable backends. Maybe instead we can at some point add a tutorial on how to do these things with differentiation in mind.

hcipy/field/util.py
```diff
@@ -52,44 +55,71 @@ def _find_common_size(*args):
...
+    center : scalar or array_like
+        The center point. The grid will by symmetric around this point. Can be an xp array for differentiability.
     has_center : boolean
         Does the grid has to have the center as one of its points. If this is
         False, this does not mean that the grid will not have the center.
+    xp : module, optional
+        The array namespace to use. If not provided, will be inferred from extent/center if they are arrays,
+        otherwise defaults to numpy.
```
MatthijsMars: I don't think this is true, if dims, extent, center are all scalar and we use new_style_fields it will give an error, see comment on `infer_xp`, same for all other grid generators

hcipy/_math/backends.py
```diff
@@ -0,0 +1,212 @@
...
+    if len(arrays) > 0:
+        try:
+            return array_api_compat.array_namespace(*arrays)
+        except (ValueError, TypeError):
+            pass
+
+    # If we get here, xp could not be inferred
+    if Configuration().core.use_new_style_fields:
+        raise ValueError("xp must be specified when arrays don't provide it")
+    return np
```
MatthijsMars, line 37: In the grid helper functions you mention that it defaults to numpy if xp is not defined or not inferred from the array. However with only scalars/None, if  the config specifies new_style_fields then this will give a value error. Is this the expected behaviour? Then the comments in the helper functions is not always true. We should maybe also test these cases to make sure they behave as expected

hcipy/field/coordinates.py
```diff
@@ -143,16 +200,25 @@ def ndim(self):
...
+    xp : module, optional
+        The array namespace (e.g., numpy, cupy, jax.numpy). If not provided,
+        it will be inferred from the input arrays.
     '''
-    def __init__(self, coords):
-        self.coords = [np.array(c) for c in coords]
+    def __init__(self, coords, xp=None):
+        if xp is None:
+            xp = infer_xp(*coords)
+        self._xp = xp
```
MatthijsMars, line 220: In these classes you use `self._xp` and in grids you use `self.xp`. Is there a particular reason for this? we should probably keep it consistent across the codebase as it will already be confusing for non-expert users.

hcipy/field/coordinates.py
```diff
@@ -143,16 +200,25 @@ def ndim(self):
...
         A tuple of a list of positions for each dimension.
+    xp : module, optional
+        The array namespace (e.g., numpy, cupy, jax.numpy). If not provided,
+        it will be inferred from the input arrays.
     '''
-    def __init__(self, coords):
-        self.coords = [np.array(c) for c in coords]
+    def __init__(self, coords, xp=None):
+        if xp is None:
+            xp = infer_xp(*coords)
```
MatthijsMars, line 219: Here you'll have the same problem as with the grid helper functions where if coords are just lists, with new_style_fields it will just give an error and not fall back to numpy. We should have a consistent fallback plan / documentation. Either fallback to numpy or always give error.

tests/test_math.py
```diff
@@ -136,3 +138,98 @@ def test_random_generator_different_sizes(xp, distribution):
...
+    ((3, 5), -1, False),
+    ((2, 3, 4), (0, 2), False),
+    ((2, 3, 4), None, False),
+    ((2, 3, 4), 1, True),
+    ((4, 6), (0, 1), True),
+]
+
+def _check_median(xp, shape, axis, keepdims, *, inject_nans=False):
+    rng = np.random.default_rng(seed=0)
+    x_np = rng.standard_normal(shape)
```
MatthijsMars, line 227: shouldn't we write these tests for an arbitrary backend so we can test array_API compatibility and in the future other backend capability? Currently the _torch_median is also not tested.

MatthijsMars (commented):
Overall the implementation seems good with only a few minor comments. One of them is how we deal with not specifically giving an xp when using the new-style fields. Documentation says it should fall back to numpy but it only does with the old-style fields. If I run the tests with the new-style-field flag turned on the field tests all fail:

```
====================================== short test summary info =======================================
FAILED tests/test_field.py::test_field_dot - ValueError: xp must be specified when arrays don't provide it
FAILED tests/test_field.py::test_field_trace - ValueError: xp must be specified when arrays don't provide it
FAILED tests/test_field.py::test_field_inv - ValueError: xp must be specified when arrays don't provide it
FAILED tests/test_field.py::test_field_transpose - ValueError: xp must be specified when arrays don't provide it
FAILED tests/test_field.py::test_field_conjugate_transpose - ValueError: xp must be specified when arrays don't provide it
FAILED tests/test_field.py::test_field_adjoint - ValueError: xp must be specified when arrays don't provide it
FAILED tests/test_field.py::test_field_inverse_tikhonov - ValueError: xp must be specified when arrays don't provide it
FAILED tests/test_field.py::test_field_inverse_truncated - ValueError: xp must be specified when arrays don't provide it
FAILED tests/test_field.py::test_field_inverse_truncated_modal - ValueError: xp must be specified when arrays don't provide it
FAILED tests/test_field.py::test_field_cross - ValueError: xp must be specified when arrays don't provide it
FAILED tests/test_field.py::test_field_svd - ValueError: xp must be specified when arrays don't provide it
FAILED tests/test_field.py::test_grid_hashing_and_comparison - ValueError: xp must be specified when arrays don't provide it
FAILED tests/test_field.py::test_grid_supersampled - ValueError: xp must be specified when arrays don't provide it
FAILED tests/test_field.py::test_field_arithmetic[NewStyleField] - ValueError: xp must be specified when arrays don't provide it
FAILED tests/test_field.py::test_field_arithmetic[OldStyleField] - ValueError: xp must be specified when arrays don't provide it
FAILED tests/test_field.py::test_field_pickle[NewStyleField] - ValueError: xp must be specified when arrays don't provide it
FAILED tests/test_field.py::test_field_pickle[OldStyleField] - ValueError: xp must be specified when arrays don't provide it
========================================= 17 failed in 0.84s =========================================
```

This is because in `test_field.py` you only test with a parametrize that gives the old and newstyle field objects, while in `_math/backends.py` we check for the flag in the configuration and this behaviour is not captured in the actual tests. In `test_field.py` we should probably specify the xp for newstylefields but this would need a refactor of the file. 

Are there more places where we compare against the config file but then test by passing in the old/newstyle field objects? Also, is generally the newstyle field mostly still untested? I think we should make sure the tests pass for both the old and newstyle fields (at least for the parts that are finished) using the array api and numpy backend.

tests/test_math.py
ehpor: This test generates a Numpy array, then converts it into a xp array using the xp fixture. It then performs the median operation on both (using Numpy for the Numpy array and xp for the xp array). It then compares those results in an assert. It therefore tests with arbitrary backends and tests full array API compatibility of the `median()` and `nanmedian()` functions.

The only issue I can see is the use of `np.asarray(result)`, which, while working as expected for the backends currently tested, it is not general and could use our `to_numpy()` function for consolidation.

hcipy/field/util.py
ehpor: Totally true. AI kept putting this back in, even though I told it to stop, and I just gave up. Forgot to fix this before committing.

hcipy/field/util.py
ehpor: Agreed.

hcipy/_math/backends.py
ehpor: I'll fix the docstrings in the helper functions (and anywhere else this is used).

Testing of this currently optional functionality will be performed in a separate testing PR.

hcipy/field/coordinates.py
ehpor: The reason is read-only properties. Users of the class always use `coords.xp` and `grid.xp`. Both of these have a property `xp` but only coordinates store the backend. The backend should be read-only: I should match the backend of the internal arrays. So the backend is stored in a private variable `Coords._xp` and made public via the properties on both `Coords.xp` and `Grid.xp`.

hcipy/field/coordinates.py
ehpor: Agreed. But disagree on the Numpy fallback always, as we discussed offline.

hcipy/field/util.py
ehpor: Fixed by d02f081079468b2deaa2c671caf392eb29751fd2.

hcipy/field/util.py
ehpor: Fixed by e0194d98249ad3f0f9c978af422fe920fd3cc0ad.

hcipy/_math/backends.py
ehpor: Docstrings of helper functions fixed by e0194d98249ad3f0f9c978af422fe920fd3cc0ad.

ehpor:
@MatthijsMars Ready for re-review.

tests/test_math.py
MatthijsMars: Okay I missed that. That makes sense

MatthijsMars (commented):
Changes look good to me and you've cleared up the other points. What in my opinion still remains is this problem:


> Overall the implementation seems good with only a few minor comments. One of them is how we deal with not specifically giving an xp when using the new-style fields. Documentation says it should fall back to numpy but it only does with the old-style fields. If I run the tests with the new-style-field flag turned on the field tests all fail:
> 
> ```
> ====================================== short test summary info =======================================
> FAILED tests/test_field.py::test_field_dot - ValueError: xp must be specified when arrays don't provide it
> FAILED tests/test_field.py::test_field_trace - ValueError: xp must be specified when arrays don't provide it
> FAILED tests/test_field.py::test_field_inv - ValueError: xp must be specified when arrays don't provide it
> FAILED tests/test_field.py::test_field_transpose - ValueError: xp must be specified when arrays don't provide it
> FAILED tests/test_field.py::test_field_conjugate_transpose - ValueError: xp must be specified when arrays don't provide it
> FAILED tests/test_field.py::test_field_adjoint - ValueError: xp must be specified when arrays don't provide it
> FAILED tests/test_field.py::test_field_inverse_tikhonov - ValueError: xp must be specified when arrays don't provide it
> FAILED tests/test_field.py::test_field_inverse_truncated - ValueError: xp must be specified when arrays don't provide it
> FAILED tests/test_field.py::test_field_inverse_truncated_modal - ValueError: xp must be specified when arrays don't provide it
> FAILED tests/test_field.py::test_field_cross - ValueError: xp must be specified when arrays don't provide it
> FAILED tests/test_field.py::test_field_svd - ValueError: xp must be specified when arrays don't provide it
> FAILED tests/test_field.py::test_grid_hashing_and_comparison - ValueError: xp must be specified when arrays don't provide it
> FAILED tests/test_field.py::test_grid_supersampled - ValueError: xp must be specified when arrays don't provide it
> FAILED tests/test_field.py::test_field_arithmetic[NewStyleField] - ValueError: xp must be specified when arrays don't provide it
> FAILED tests/test_field.py::test_field_arithmetic[OldStyleField] - ValueError: xp must be specified when arrays don't provide it
> FAILED tests/test_field.py::test_field_pickle[NewStyleField] - ValueError: xp must be specified when arrays don't provide it
> FAILED tests/test_field.py::test_field_pickle[OldStyleField] - ValueError: xp must be specified when arrays don't provide it
> ========================================= 17 failed in 0.84s =========================================
> ```
> 
> This is because in `test_field.py` you only test with a parametrize that gives the old and newstyle field objects, while in `_math/backends.py` we check for the flag in the configuration and this behaviour is not captured in the actual tests. In `test_field.py` we should probably specify the xp for newstylefields but this would need a refactor of the file.
> 
> Are there more places where we compare against the config file but then test by passing in the old/newstyle field objects? Also, is generally the newstyle field mostly still untested? I think we should make sure the tests pass for both the old and newstyle fields (at least for the parts that are finished) using the array api and numpy backend.


These are an indication that what we're testing is not actually how the user will experience these functions (we test by passing NewStyleField implicitly, not by using the config flag which is used for this xp valueError). These are errors that are introduced by making changes to the grids and we have already made fields Array API compliant before, so these tests should pass under the right conditions.   I would prefer to fix it here now, but at the least there should be an issue underpinning this specific untested scenario.

ehpor:
I'll again clarify the context around testing. This PR does not break the default non-experimental test suite. However, when the _experimental_ feature flag is enabled, the suite currently (on master) has ~1,700 out of ~2,000 failures. This PR actually improves that baseline: 27 tests now pass in experimental mode (in `test_grid.py`) that previously failed. Simultaneously, 5 new failures appear in `test_field.py`. Note: that test file has never fully passed under experimental mode. The net result is +22 passing tests overall. Addressing these 5 in isolation wouldn't meaningfully advance the test suite in experimental mode given the broader refactor scope.

Since Array API support is an _experimental_ feature hidden behind a feature flag, full test parity cannot yet be expected. This PR does not break the default non-experimental CI. Test failures in experimental mode will be fixed and addressed systematically as tracked in #335.

I'll address the merge conflicts now.

ehpor:
@MatthijsMars back to you.

MatthijsMars (dismissed):
Thanks for the clarification, with that in mind, this looks good to me

ehpor:
Fixed the test failures due to merging of #330 (specifically the addition of `array_api_strict` to the `xp` fixture.

@MatthijsMars Needs re-approval.

MatthijsMars (approved):

PR #306, opened 2026-02-25
Add HWO EAC-1 aperture

GeorgeVassilakis:
Hey Emiel,

I thought I'd add the pupil/aperture for HWO's EAC-1 since I saw that EAC-2 was already integrated, and I had the parameters for EAC-1 anyways (from the HWO Working Group GitHub, linked [here](https://github.com/HWO-GOMAP-Working-Groups/Sci-Eng-Interface/blob/main/hwo_sci_eng/obs_config/Tel/EAC1.yaml)). It would make my codebase cleaner if I just called the HCIPy `make_***_aperture` function! 

There is an inconsistency between the pupil diameter defined in the configuration file (7.225765 m) and the inferred diameter created from the hexagon size + gap size (point-to-point size = 1.65m and optical gap = 0.006m creates a 7.1687m mirror), so I chose to force the pupil diameter with a rescale of the components within it. I made note of this in the docstring as well. If you think we should define the diameter based on the diameter inferred from the segments + gap, let me know, and I'll change it.

Apologies for the merging of new `master` commits into this branch. I had made this branch in my `GeorgeVassilakis/hcipy` repo last month and forgot to actually make this PR.


ehpor:
Great PR and very appreciated.

The code looks good to me, no complaints there. The documentation is in order too.

I would like to get to the bottom of the scaling factor before merging though. I thought about a projection effect, where the actual segments are the correct size but they are tilted in 3D space and therefore appear slightly smaller. But that would make the pupil diameter of this version _larger_ than the actual pupil diameter. To be clear: a scaling factor seems totally reasonable to me, but I'd like to know the reason for the discrepancy before blindly accepting a scaling factor.

GeorgeVassilakis:
Got it, I'll check with some members of the HWO team and get back to you!

GeorgeVassilakis:
Hi Emiel! Just saw you re-assigned this PR, apologies for the very delayed follow-up. I checked with colleagues on the HWO side, but I wasn’t able to establish the origin of the discrepancy between the quoted pupil diameter and the segment/gap geometry. I also don’t expect we’ll have a definitive clarification in the near term.

In the absence of a justification for the scale factor, would you prefer that I rewrite the implementation to preserve the quoted segment size and gap, which would then imply the approximately 7.1687 m outer diameter, and just document the conflict with the separately quoted 7.225765 m value? Might as well just choose one and document the discrepancy at this point!

ehpor:
@GeorgeVassilakis Yes, let's add a `.. warning::` to the docstring. It should just say that the outer diameter doesn't match the quoted outer diameter of xxxx meters, and that we're not quite sure why. Use at your own risk.

Afterwards we can merge this.

GeorgeVassilakis:
Excellent, I'll take care of it now and let you know when it's done.

GeorgeVassilakis:
@ehpor, should be all set now! Code uses the segment size + optical gap directly which results in a diameter of ~7.1687m. Documentation includes the `.. warning::` too! Lmk if there's anything else you want done on this PR.

ehpor:
I'll rebase and remove the unnecessary merge commits. Then I'll approve and merge.

ehpor:
Rebased without issues as expected.

ehpor (approved):
LGTM. Issue with pupil diameter noted in docstring as warning.

ehpor:
Thanks @GeorgeVassilakis 🎉

GeorgeVassilakis:
No problem, thanks for merging it in @ehpor! Fair warning: I may hit you with another PR later this week. I am working on an orthonormal mode basis for EAC2's keystone segments, analogous to the Hexikes that I integrated a while back. I have a prototype but am validating and trimming before I actually send it your way. If I'm correct, keystone segments in EAC2 can only go through piston/tip/tilt, with nothing higher order. Susan Redmond (@sfr348) and I discussed this a while back, and I think it's now potentially worth putting in the codebase!

PR #308, opened 2026-03-03
Fix `SimpleVibration` class.

### LINKED ISSUE #307: Bug in SimpleVibration class
I found a bug in the SimpleVibration class, something is awry with the assignment of `frequency`. 

From Emiel: setting line 9 in [`vibration.py`](https://github.com/ehpor/hcipy/blob/master/hcipy/optics/vibration.py) to `self._frequency = frequency` is a quick fix. 

Example error:
Image

ehpor:
This PR fixes #307, adds a test to confirm and adds docstrings to all functions.

### DIFF
--- a/hcipy/optics/vibration.py
+++ b/hcipy/optics/vibration.py
@@ -3,6 +3,25 @@ import numpy as np
 from .optical_element import OpticalElement
 
 class SimpleVibration(OpticalElement):
+    '''A simple vibration model.
+
+    Parameters
+    ----------
+    mode : Field
+        The spatial mode of the vibration.
+    amplitude : scalar
+        The amplitude of the vibration. After multiplying the mode and
+        the amplitude, the resuilting units should be meters.
+    frequency : scalar
+        The temporal frequency of the vibration in Hz.
+    phase_0 : scalar, optional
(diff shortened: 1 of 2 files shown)


hcipy/optics/vibration.py
```diff
@@ -3,34 +3,83 @@
 from .optical_element import OpticalElement
 
 class SimpleVibration(OpticalElement):
+    '''A simple vibration model.
+
+    Parameters
+    ----------
+    mode : Field
+        The spatial mode of the vibration.
```
julesfowler, line 15: If I were looking at these docs fresh, an example like "for example vibrational jitter via tip or tilt" would be a helpful addition here. But I think it is probably clear to most as is if that feels too wordy for a docstring.

julesfowler (dismissed):
Changes look good! I tested the local branch on my machine and it seems to be working. I left you one comment in line, but not a required change for this to be merged, just a suggestion.

hcipy/optics/vibration.py
ehpor: Fixed in 2d4a7f0.

julesfowler (approved):

PR #309, opened 2026-03-03
Add state space dynamics for time-varying optical aberrations.

ehpor:
This PR introduces classes for modeling time-varying optical aberrations using continuous-time state space dynamics. The implementation includes a `StateSpaceDynamics` class in `hcipy/util/stochastic_dynamics.py` that generates trajectories for stochastic processes of the form dx/dt = Ax + Bw, where A is the transition matrix, B is the noise matrix, and w is white noise. The class uses exact matrix exponential for deterministic evolution and exact noise covariance derived from the continuous Lyapunov equation for the stochastic component.

The PR also includes several helper functions for creating state space models: `ar_to_state_space()` converts discrete-time AR coefficients to continuous-time state space, `make_continuous_time_companion_matrix()` creates companion form matrices from poles, and `make_random_state_space` generates stable random transition matrices. A new `DynamicSurfaceAberration` class was added to `hcipy/optics/aberration.py`. This class wraps a `StateSpaceDynamics` object and applies phase modulation to wavefronts based on the current state. It accepts a `refractive_index` parameter (scalar or callable) to convert surface height to phase. The PR also adds a `DampedHarmonicVibration` class which uses composition with `DynamicSurfaceAberration`. This simulates a damped harmonic oscillator, vibrating a certain wavefront mode.

Tests were added to verify the stationary distribution, explicit initial states, backwards evolution blocking, long-term stationarity, AR conversion, and the refractive index parameter.


hcipy/optics/aberration.py
```diff
@@ -178,3 +178,129 @@ def backward(self, wavefront):
         wf = self.fresnel.forward(wavefront)
         wf = self.surface_aberration.backward(wf)
         return self.fresnel.backward(wf)
+
+class DynamicSurfaceAberration(OpticalElement):
```
syhaffert, line 182: Shouldn't this class derive from SurfaceApodizer? That already implements a lot of the necessary code. The only difference is that this class has a dynamic surface_sag.

hcipy/optics/vibration.py
```diff
@@ -88,3 +91,183 @@ def backward(self, wavefront):
 
         wf.electric_field *= np.exp(-1j * (self.mode * self.amplitude / wf.wavelength * np.sin(self.phase)))
         return wf
+
+class DampedHarmonicVibration(OpticalElement):
```
syhaffert: Why make this a separate class? This should derive from the DynamicSurfaceAberration. Now you need to reimplement the interfaces again to expose access to the evolve_until etc..

hcipy/util/stochastic_dynamics.py
```diff
@@ -0,0 +1,319 @@
...
+        """
+        if t < self._t:
+            raise ValueError(f"Backwards evolution not allowed: target time {t} is less than current time {self._t}")
+
+        if t == self._t:
+            return
+
+        dt = t - self._t
+
+        A_disc = expm(self.transition_matrix * dt)
```
syhaffert, line 149: Isn't it better to make a separate property. Then we can cache this calculation because doing eigenvalue decomposition is quite expensive. We can speed it up substantially since people usually only evolve with uniform time steps.

hcipy/optics/aberration.py
ehpor: This was written _WAYYY_ before the apodizer refactor, so back then the interfaces were not compatible. I can refactor this now.

hcipy/optics/vibration.py
ehpor: Agreed. Fixed in 54b95bf07b5a7d83f2e7a0ebab95e45cc192dd35.

hcipy/optics/aberration.py
ehpor: Disagree. The `SurfaceApodizer` class has a `surface_sag` property that allows both getting and setting. The `DynamicSurfaceAberration` does not allow _setting_ of the `surface_sag`, just getting, and therefore is strictly speaking _not_ a `SurfaceApodizer`: it violates the Liskov substitution principle.

Instead I refactored to use composition with an internal `SurfaceApodizer` object, which eliminates most of the duplicated code.

hcipy/util/stochastic_dynamics.py
ehpor: In principle I agree. However, my use case is different. I use very different time steps each iteration. Also for low dimensionality, it doesn't really matter. 

| n    | `evolve_until()` (uncached) | `evolve_until()` (cached) | sag calculation (512x512) |
|------|-------------------------|---------------------|--------------------|
| 2    | 12 µs                   | 1.8 µs              | 122 µs  |
| 4    | 14 µs                   | 1.8 µs              | 187 µs             |
| 8    | 20 µs                   | 1.9 µs              | 267 µs             |
| 16   | 33 µs                   | 1.9 µs              | 427 µs             |
| 32   | 67 µs                   | 2.1 µs              | 747 µs             |
| 64   | 240 µs                  | 2.8 µs              | 1.39 ms            |
| 128  | 843 µs                  | 5.3 µs              | 2.67 ms            |
| 256  | 4.75 ms                 | 11.7 µs             | 5.15 ms            |
| 512  | 25.9 ms                 | 24.5 µs             | 10.2 ms            |
| 1024 | 158 ms                  | 164 µs              | 20.3 ms            |
| 2048 | 1220 ms                 | 1.11 ms             | 40.5 ms            |
| 4096 | 11.9 s                  | 4.56 ms             | 81 ms              |

I'd say that for `n=256 it starts to matter, before I don't really care.

Finally, caching is difficult to set up correctly:
* raw float `dt` is almost never completely the same.
* rounded floats don't suffer from that, but you need to find a rounding level that does not smush together neighboring `dt`s, and that level depends on the bit depth
* caching wouldn't work well with Jax, I don't think.

I could speed things up using a Schur decomposition instead of eigenvalue decomposition, which numerically is better too, but that's for-loop heavy code and I also don't know how Jax will respond to that. I can also work in the eigenbasis of the transition matrix. Let me work on that a little bit.

PR #323, opened 2026-06-06
Add fourier convolution and correlation functions.

ehpor:
## Summary

Adds `make_fourier_convolution()` and `make_fourier_correlation()`, which construct a `FourierFilter` pre-computed from a spatial kernel (provided as a Field or a Field generator). Also adds a private `_make_fourier_kernel_filter()` helper to share the common logic. Includes tests against a reference implementation for various grid sizes and zeropadding factors, plus error-case tests for delta mismatches and oversized kernels.

## Usage

```python
import numpy as np
from hcipy import *

grid = make_pupil_grid(128, 1)
noise = Field(np.random.normal(0, 1, grid.size), grid)

def gaussian_kernel(grid, sigma=0.02):
    r = grid.as_('polar').r
    return Field(np.exp(-r**2 / (2 * sigma**2)), grid)

conv = make_fourier_convolution(grid, gaussian_kernel, q=2)
smoothed = conv.forward(noise)

plt.figure(figsize=(10, 4))
plt.subplot(121)
imshow_field(noise, cmap='RdBu')
plt.title('Gaussian noise')

plt.subplot(122)
imshow_field(smoothed.real, cmap='RdBu')
plt.title('Noise convolved with Gaussian')

plt.tight_layout()
plt.show()
```

convolution_demo

### DIFF
--- a/hcipy/fourier/__init__.py
+++ b/hcipy/fourier/__init__.py
@@ -7,6 +7,8 @@ __all__ = [
     'is_fft_grid',
     'FastFourierTransform',
     'FourierFilter',
+    'make_fourier_convolution',
+    'make_fourier_correlation',
     'FourierShift',
     'FourierShear',
     'FourierRotation',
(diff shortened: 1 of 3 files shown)


syhaffert (approved):
Approved for merging.

PR #324, opened 2026-06-08
Remove deprecated functions, classes, properties and parameters.

ehpor:
All of these were deprecated over three years ago.

* Aperture generation functions without `make_` prefix.
* The `SingleModeFiber` class (replaced with `SingleModeFiberInjection`).
* The `make_agnostic_optical_element()` function (replaced with the `AgnosticOpticalElement` class).
* The `set_color_scheme()` function (no replacement; leftover from convenience code from when I created HCIPy).
* The `M` property and parameter on the `SVD` class (replaced with `matrix`).


syhaffert (approved):
Looks good to me. I agree that it's time to remove some of the deprecated functions.

PR #328, opened 2026-06-09
Add tests for `GaussianBeam` and `DynamicOpticalSystem`.

ehpor:
This improves test coverage of these lesser-used classes.

Also fixes a bug in `DynamicOpticalSystem` where `evolve_until()` crashed when there were no callbacks registered.


raphaelpclt (approved):
Looks good

PR #329, opened 2026-06-09
Do not ignore overwrite parameter in `write_fits()`.

ehpor:
The `overwrite` was hardcoded as `True` and the function parameter was completely ignored.


raphaelpclt (approved):
Ok!

PR #330, opened 2026-06-09
Add more distributions and functions on `RandomGenerator`.

ehpor:
This PR adds the `uniform()`, `exponential()` and `choice()` functions that are available for Numpy into the `RandomGenerator` classes. It also adds `array_api_strict` as a backend, allowing for testing with that backend. Tests have been added for all functions.

This now should be all random functions that are used internally in HCIPy.

### DIFF
--- a/tests/test_math.py
+++ b/tests/test_math.py
@@ -74,6 +74,7 @@ def test_random_generator_poisson(xp, lam):
     # Get samples from the Poisson distribution.
     rng = make_random_generator(xp, seed=42)
     samples = rng.poisson(lam=lam, size=(10000,))
+    samples = xp.astype(samples, xp.float32)
 
     # Basic statistical tests for poisson distribution
     assert np.isclose(float(xp.mean(samples)), lam, atol=0.1)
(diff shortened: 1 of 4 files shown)


tests/test_math.py
```diff
@@ -91,48 +92,117 @@ def test_random_generator_gamma(xp, scale, shape):
...
+@pytest.mark.parametrize('low, high', ((0.0, 1.0), (-2.0, 5.0)))
+def test_random_generator_uniform(xp, low, high):
+    # Get samples from the Uniform distribution.
+    rng = make_random_generator(xp, seed=42)
+    samples = rng.uniform(low, high, size=(10000,))
+
+    # Basic statistical tests for uniform distribution
+    mean = (low + high) / 2
+    std = (high - low) / math.sqrt(12)
+    assert np.isclose(float(xp.mean(samples)), mean, atol=0.05)
```
MatthijsMars, line 104: Should we replace all np.isclose /np.allclose with a backend independent one as introduced in the grids PR?

ehpor: In principle yes. The backend-independent `all_close()` still requires Arrays as input, which in this case `mean` is not. But that can be fixed. However, since #305 has not yet been merged, I cannot use/fix that function here.

ehpor:
Fixed the merge conflict. Ready for re-review. @MatthijsMars

tests/test_math.py
MatthijsMars: Okay we should probably add that as a todo somewhere then

MatthijsMars (approved):
looks good to me

PR #332, opened 2026-06-20
Fix number of core overrides.

### LINKED ISSUE #229: User Settable Number of Cores
As far as I can tell, there isn't a way for the user to set the number of cores used in the FFT.

I've tried using `os.sched_setaffinity` but this isn't a great way of doing this because you can only set specific cores. This overrides the CPU scheduler and ends up being very slow and doesn't scale very well at all unless you use all the cores on the system.

My particular use case is that I'm running HCIPy on a shared use server with a bunch of cores, and I need to be able to limit the number of threads used by my processes.

Probably need to change the [file here.](https://github.com/ehpor/hcipy/blob/3cdb57f4d5c314dd1589fcd1cc6a669c82b9d39f/hcipy/_math/cpu.py#L3)

ehpor:
We should use BLAS number of core overrides (environment variables), HPC scheduler number of cores (environment variables), the number of cores BLAS decided to use (threadpoolctl) before resorting to maxing out the core count.

This PR adds `threadpoolctl` as a new dependency. This is a widely-used package, eg. a dependency of scikit-learn.

Fixes #229.

### DIFF
--- a/hcipy/_math/cpu.py
+++ b/hcipy/_math/cpu.py
@@ -1,25 +1,50 @@
 import os
 import platform
 import multiprocessing
+import threadpoolctl
 
 def get_num_available_cores():
-    '''Get the number of cores available on the system.
+    '''Get the number of cores available to the current process.
 
-    The attempt at retrieving the number of cores that our process
-    is assigned may not be available on all operating systems. On
-    unsupported operating systems, the total number of cores is
-    returned instead.
+    Checks, in order of priority:
+    1. Actual BLAS thread count via threadpoolctl (if available and loaded)
+    2. BLAS/threading environment variables (OMP_NUM_THREADS, etc.)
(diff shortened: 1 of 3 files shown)


raphaelpclt (dismissed):
Looks good

ehpor:
@raphaelpclt On second thought, I just looked at the codecoverage and it's pretty bad. This is not trivial code, so I'm holding off on merging until that is resolved.

ehpor:
I factored out the specific number of core retrieval functions into their own separate functions and tested them separately. I also added specific tests for the cgroup functions, which are a bit more specific.

@raphaelpclt Ready for review.

ehpor:
Fixed merge conflict.

raphaelpclt (approved):
Ok

PR #336, opened 2026-06-30
Refactor apodizer classes.

ehpor:
This PR rewrites the apodization class hierarchy to inherit from `OpticalElement` instead of `AgnosticOpticalElement`, eliminating all caching and grid-agnostic parameter evaluation. Dispersion optics classes (`TiltElement`, `ThinPrism`, `Prism`, `PhaseGrating`) also become standalone `OpticalElement` subclasses that compute sag/phase on-the-fly from the wavefront's grid, removing the grid constructor parameter.

This reduction in caching simplifies the transition to autodiff frameworks. Those need to know the entire calculation from start to finish to do differentiation, and caching of intermediate values limits that.

Several bugs were fixed in addition to the refactor:
1. `ThinPrism.minimal_deviation_angle()` referenced `self.prism_angle` which was never set.
2. `PhaseGrating.amplitude` property getter returned `self._amplitude` which was never set.
3. `MultiplexedComplexSurfaceApodizer.backward()` used wrong mathematical operation (division by amplitude mask rather than multiplication by its conjugate.
4. Missing docstring of `ComplexSurfaceApodizer` and `MultiplexedComplexSurfaceApodizer` classes.
5. Documentation typo ("sacalar") in the `Prism` class docstring.

### DIFF
--- a/doc/tutorial_notebooks/ElectricFieldConjugation/ElectricFieldConjugation.ipynb
+++ b/doc/tutorial_notebooks/ElectricFieldConjugation/ElectricFieldConjugation.ipynb
@@ -307,7 +307,7 @@
    "source": [
     "aberration_distance = 100e-3 # m\n",
     "\n",
-    "aberration_at_distance = SurfaceAberrationAtDistance(aberration, aberration_distance)\n",
+    "aberration_at_distance = SurfaceAberrationAtDistance(pupil_grid, aberration, aberration_distance)\n",
     "\n",
     "aberrated_pupil = aberration_at_distance(Wavefront(aperture, wavelength))\n",
     "\n",
--- a/hcipy/optics/aberration.py
+++ b/hcipy/optics/aberration.py
@@ -131,13 +131,15 @@ class SurfaceAberrationAtDistance(OpticalElement):
 
     Parameters
     ----------
+    input_grid : Grid
+        The grid on which the surface aberration operates.
     surface_aberration : OpticalElement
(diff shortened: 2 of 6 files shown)


syhaffert (approved):
This is a good PR. I think the AgnosticOpticalElement was quite over engineered. This will make it easier to implement new elements, and it is clearer for new users. It will also make it easier to access the actual apodization surfaces, which I sometimes struggled with in the past. This looks ready to be merged. I approve it.

raphaelpclt (commented):
I also feel the simplification is very welcome, and the code looks good. 

As a minor side note: are there impacts on performance people should expect from not caching the instance data anymore? I feel it's ok but I am not sure to grasp all of what the AgnosticOpitcalElement was caching.

syhaffert:
It will slow down the optical elements by a bit because its not caching anymore. It now needs to generate the intermediate transfer functions / apodizers on the fly. That is necessary for compatibility with JAX.

ehpor:
There will be slowdowns in cases where users gave a Field generator or a function of wavelength to any of the apodizer classes. Before, the Field generator was called once and the result was cached. Now, the Field generator is called on each forward and backward propagation. If users were doing a lot of work in that Field generator, they'll see slowdown.

In cases where a Field was given directly, no new calculations are done (afaik, I might've missed a multiplication by the wavenumber somewhere that was cached before but the vast majority of those were calculated from scratch before as well).

PR #337, opened 2026-06-30
Misc bugfixes.

ehpor:
This PR fixes a variety of bugs:

* Typos in variable names.
* Docstrings of two functions.
* Default parameters that made the class broken.
* Missing assertion for even `PerfectCoronagraph` order.
* Unnecessary package dependencies.
* Missing package dependency. 
* Supported grids for `NaiveFourierTransform`.
* Grid weights during shifting of polar grids.
* Dead code.
* Warnings emitted due to divide by zero (which was fixed already).
* Missing `super().__init__()` call in `SegmentedDeformableMirror`.
* Broken string formatting for error message.
* Deprecation of `ColorMap.set_bad()` from Matplotlib.

These bugs were found with the help of DeepSeek V4 (alongside maybe double as many false positives).

### DIFF
--- a/hcipy/atmosphere/atmospheric_model.py
+++ b/hcipy/atmosphere/atmospheric_model.py
@@ -43,7 +43,7 @@ class AtmosphericLayer(OpticalElement):
     height : scalar
         The height of the atmospheric layer above the ground.
     '''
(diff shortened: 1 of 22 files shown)


raphaelpclt (approved):
Looks good

PR #338, opened 2026-07-01
Add tests for microlens array classes.

ehpor:
This PR adds unit tests for microlens array classes (`MicroLensArray`, `SphericalMicroLensArray`,
`EvenAsphereMicroLensArray`). It checks roundtrip propagation (which is guaranteed to be identity as MLAs are phase-only), attribute access and extent of the MLAs.

It also makes `closest_points()` a private helper rather than public. A deprecation warning has been added and the old function still works for the time being.

### DIFF
--- a/tests/test_micro_lens_array.py
+++ b/tests/test_micro_lens_array.py
@@ -66,8 +66,11 @@ def test_micro_lens_array_multiple_lenslets():
     assert np.all(mla.mla_opd[~assigned] == 0)
 
     wf = Wavefront(input_grid.ones())
-    mla.forward(wf)
-    mla.backward(wf)
+    wf_fwd = mla.forward(wf)
+    wf_bwd = mla.backward(wf_fwd)
+
+    # Make sure that no amplitude is applied.
+    assert np.allclose(wf.electric_field, wf_bwd.electric_field)
 
 def test_spherical_micro_lens_array():
     input_grid = make_pupil_grid(64, 1)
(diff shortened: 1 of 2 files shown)


tests/test_micro_lens_array.py
```diff
@@ -0,0 +1,143 @@
...
+    mla = MicroLensArray(input_grid, lenslet_grid, focal_length, lenslet_shape=lenslet_shape)
+
+    assert mla.mla_index.size == input_grid.size
+    assigned = mla.mla_index >= 0
+    assert np.any(assigned)
+    assert np.all(mla.mla_index[assigned] < 9)
+    assert np.all(mla.mla_opd[assigned] < 0)
+    assert np.all(mla.mla_opd[~assigned] == 0)
+
+    wf = Wavefront(input_grid.ones())
```
syhaffert, line 68: Waht do these lines 68-70 do? There is no assert or check on anything.

syhaffert (commented):
Only 1 comment. After that its ready for merging.

tests/test_micro_lens_array.py
ehpor: They just check that things run. I honestly could ensure that they only apply phase (ie. roundtrip = identity). Let me do that.

ehpor: Fixed in e7a409ce54431c9bf72ecbaab41080ffb43890e5.

syhaffert (approved):

PR #339, opened 2026-07-07
Add CI for type checking.

ehpor:
This PR adds the CI infrastructure for type hint checking using mypy, along with a GitHub problem matcher so that errors are annotated inline in the PR diff.

The actual work of adding type hints to the codebase will be done in subsequent PRs. This PR is purely about setting up the tooling:
* the CI workflow (.github/workflows/typing.yml)
* the problem matcher (.github/mypy_problem_matcher.json)
* the mypy configuration.

Future PRs with type annotations are automatically validated.

### DIFF
--- a/hcipy/field/field.py
+++ b/hcipy/field/field.py
@@ -1,6 +1,7 @@
 import numpy as np
 from ..config import Configuration
 from .._math import fft as hcipy_fft
+from typing import Any
 
 class FieldBase:
     '''The value of some physical quantity for each point in some coordinate system.
@@ -961,9 +962,9 @@ def _make_namespace(slots):
     return Namespace()
 
 # Global cache for FieldNamespace instances.
-_field_namespace_cache = {}
-_last_backend = None
-_last_field_namespace = None
+_field_namespace_cache: dict[str, Any] = {}
+_last_backend: str | None = None
+_last_field_namespace: Any | None  = None
 
 def make_field_namespace(backend):
     """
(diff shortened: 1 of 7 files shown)


ehpor:
There are several failing things in this test.

* atmosphere submodule: these are fixed by #337.
* config, glass and deformable mirror submodules: these are fixed by #324
* apodization submodule: these are fixed by #336

Nevertheless, this PR is ready for review as the check is not mandatory yet. I will make it mandatory once all type checks pass.

hcipy/field/field.py
```diff
@@ -1,6 +1,7 @@
 import numpy as np
 from ..config import Configuration
 from .._math import fft as hcipy_fft
+from typing import Any
 
 class FieldBase:
```
raphaelpclt, line 6: Doesn't `from __future__ import annotation` need to be added here for backward compatibility since there are annotations in this file? If not, how is it different from the `coordinates.py` or the `grid.py` files?

ehpor: The import is for future references: types which are not yet known at the time of the statement. Since Python 3.5 you were able to put the type in quotes, which meant that it will be evaluated after the whole file has finished importing. Since Python 3.7 you were able to use this import and not need to quotes anymore. Since Python 3.14, the import is no longer necessary. So importing `annotations` is to allow me to not use quotes and be simply able to remove the import once we drop Python 3.13 support.

* `grid.py` uses `Grid` as a type inside the `Grid` class definition. At that point, `Grid` is not yet defined.
* `coordinates.py` uses `Coords` as a type inside the `Coords class definition. At that point, `Coords` is not yet defined.

In `field.py` we don't have such a case: we only use `Any` and standard types.

raphaelpclt: Oh ok yes types in `field.py` are already known, that's what I've missed.

raphaelpclt (approved):
I'm ok with it.

ehpor:
Waiting to merge until after #337 to ensure that that fixes all type hints (so that I can start adding type hints from a clean slate).

ehpor:
The type hinting check is green now after all dependent PRs were merged.

Merging this PR now.

PR #341, opened 2026-07-23
Disable hypothesis deadlines during testing.

ehpor:
The deadline setting exists to catch performance regressions, and prevent slow tests. However, slow systems or heavily loaded CI servers can cause false-positive timeouts. This has been happening for us as well.

This PR turns the deadline setting off.


raphaelpclt (approved):
Ok

PR #342, opened 2026-07-27
Speed up `InfiniteAtmosphericLayer._extrude()`.

ehpor:
I changed the indexing and views in the `_extrude()` function. This allows us to avoid two copies of the phase screen per extrusion. Additionally, since there are very few pixels in the stencil compared to pixels in the phase screen, this PR switches to using indices rather than booleans for the mask.

Benchmarks on my Macbook Pro M2 Max chip for a 512x512 pupil grid:

Direction | Original | Now | Speedup
-- | -- | -- | --
top | 0.59 ms | 0.19 ms | 3.1×
bottom | 0.30 ms | 0.20 ms | 1.5×
left | 0.33 ms | 0.21 ms | 1.6×
right | 0.62 ms | 0.15 ms | 4.1×
**average** | **0.46 ms** | **0.19 ms** | **2.4×**

Note that in most cases, the interpolation takes the majority of the computation time. That part is unaffected by this PR.

### DIFF
--- a/hcipy/atmosphere/infinite_atmospheric_layer.py
+++ b/hcipy/atmosphere/infinite_atmospheric_layer.py
@@ -10,6 +10,7 @@ from scipy.ndimage import affine_transform
 
 import warnings
 import copy
+import math
 
 class InfiniteAtmosphericLayer(AtmosphericLayer):
     '''An atmospheric layer that can be infinitely extended in any direction.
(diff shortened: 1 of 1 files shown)


syhaffert (commented):
This looks good to me.

syhaffert (approved):

PR #345, opened 2026-07-28
Fix incorrect interpolation docstring.

ehpor:
Interpolation is fifth-order Bezier, not bilinear as the docstring mentioned.

Reported by @ataras2, see https://github.com/LouisDesdoigts/dLux/pull/355#issuecomment-5092898721.


syhaffert (approved):
LGTM

PR #343, opened 2026-07-28
Atmosphere interpolation speedup

ehpor:
Depends on #342.

Add a `subpixel_shift()` function that performs 5th-order Bezier interpolation using separated convolution. This operation is accelerated using Numba. There also is a fallback for other Array API backends, which simply use array views and in-place addition.

PR #347, opened 2026-07-29
Add B-spline interpolation order for `InfiniteAtmosphericLayer`.

ehpor:
Dependent on #342 and #343.

This PR introduces arbitrary order B-spline interpolation using `subpixel_shift()` and thereby `InfiniteAtmosphericLayer`. This allows people to use lower-order interpolation when their atmospheric simulation does not require the fidelity. This gives a slight performance boost. Note: in most cases for predictive control simulations at high framerate, you need high-order interpolation.

The calculation of B-spline coefficients uses [Cox–de Boor recursion formula](https://en.wikipedia.org/wiki/De_Boor%27s_algorithm), for those that are curious. The pixel shifts are checked against `scipy.ndimage.affine_transform()` for orders 1-5.

Benchmarks on Macbook Pro M2 Max:

```bash
> python benchmark_atmosphere.py
Single-step evolution (order 0): 443.1 FPS (2.257 ms/frame, 1024x1024 screen)
Single-step evolution (order 1): 259.8 FPS (3.850 ms/frame, 1024x1024 screen)
Single-step evolution (order 2): 236.2 FPS (4.234 ms/frame, 1024x1024 screen)
Single-step evolution (order 3): 219.9 FPS (4.547 ms/frame, 1024x1024 screen)
Single-step evolution (order 4): 218.5 FPS (4.578 ms/frame, 1024x1024 screen)
Single-step evolution (order 5): 212.5 FPS (4.706 ms/frame, 1024x1024 screen)
```

For the reviewer: we could consider removing the `use_interpolation` parameter, deprecating it, and letting `interpolation_order` being zero determine that no interpolation is needed (effectively nearest-pixel interpolation).

### DIFF
--- a/hcipy/atmosphere/infinite_atmospheric_layer.py
+++ b/hcipy/atmosphere/infinite_atmospheric_layer.py
@@ -10,6 +10,7 @@ from scipy import linalg
 
 import copy
 import math
+import warnings
 
 class InfiniteAtmosphericLayer(AtmosphericLayer):
     '''An atmospheric layer that can be infinitely extended in any direction.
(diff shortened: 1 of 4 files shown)


hcipy/atmosphere/infinite_atmospheric_layer.py
```diff
@@ -60,11 +60,15 @@ class InfiniteAtmosphericLayer(AtmosphericLayer):
         can be unintuitive. The default value of 2 is usually sufficient in
         most instances.
     use_interpolation : boolean
```
syhaffert, line 63: I agree that use_interpolation is superfluous and that interpolation_order=0 can achieve the same thing.

syhaffert (changes requested):
Only 1 small comment. After that it should be ready for merging.

hcipy/atmosphere/infinite_atmospheric_layer.py
ehpor: Done in 4b607db2d5b1d99c60fa030d971b51bdf2b5b80c.

syhaffert (approved):

PR #349, opened 2026-07-31
Fast DFT matrix calculation

ehpor:
## Fast DFT matrix computation

This PR adds a fast, Array-API-compatible way to compute DFT matrices — the matrices `V[i,k] = exp(1j * u[k] * x[i])` used by `MatrixFourierTransform` — and switches the MFT over to it.

### New module: `hcipy/_math/fourier.py`

Two public functions, each with a NumPy fast path and an Array API fallback:

- **`dft_matrix_regular`** — for equally spaced grids (given  `x0, dx, u0, du, Nx, Nu`). The NumPy backend uses a numba parallel implementation that factorizes the matrix as `C * A[i] * B[k] * T[i+k]`, reducing complex exponentials from `O(Nx * Nu)` to `O(Nx + Nu)`; the rest are cheap complex multiplications. The other backends use the Array API.
- **`dft_matrix_separated`** — for arbitrary (separated) grids (given coordinate arrays `x`, `u`). The NumPy backend evaluates `exp(1j * x * u)` with numexpr; other backends use the Array API.

Both support `conjugate`, `transpose` and `out=`, and respect the input/requested dtype (float32 → complex64, float64 → complex128).

### `MatrixFourierTransform` uses the new functions

The matrix computation in `_compute_matrices` no longer builds `outer` products inline. Regular input/output grids now take the numba path; non-regular separated grids fall back to the numexpr-based `dft_matrix_separated` (identical behavior to before).

### Tests

- New `dft_matrix_*` tests in `tests/test_math.py` comparing against a naive NumPy implementation to numerical precision, parametrized by the `xp` fixture, including `conjugate`/`transpose` combinations, `out=` behavior, and output dtype checks.
- `tests/conftest.py` enables JAX 64-bit mode so dtype tests exercise the full precision range.

I used Deepseek V4 Pro and Flash to generate part of this PR.

### DIFF
--- a/hcipy/_math/fourier.py
+++ b/hcipy/_math/fourier.py
@@ -1,0 +1,77 @@
+import numba
+import numexpr
+import cmath
+import numpy as np
+
+from .backends import array_namespace
+from array_api_compat import is_numpy_array, is_numpy_namespace
+
+@numba.njit(parallel=True, fastmath=True, cache=True)
+def dft_matrix_separated_numba(x0, u0, dx, du, Nx, Nu, dtype, out=None):
+    alpha = du * dx / 2.0
+    C = cmath.exp(1j * u0 * x0)
+
+    A = np.empty(Nx, dtype=dtype)
+    for i in range(Nx):
+        A[i] = cmath.exp(1j * (u0 * dx * i - alpha * i * i))
+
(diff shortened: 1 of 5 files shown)


ehpor:
Fixed merge conflict.

syhaffert (approved):

PR #351, opened 2026-07-31
Use PyPI trusted publishing for releases.

ehpor:
## Summary

Pin `pypa/gh-action-pypi-publish` to a specific commit SHA instead of the floating `release/v1` tag in `.github/workflows/release.yml`.

The release workflow relies on trusted publishing (OpenID Connect): it requests `id-token: write` and publishes to PyPI with no stored API token or password. With trusted publishing, PyPI verifies the token-exchange request comes from this repository's GitHub Actions environment, so no secrets are involved in the workflow at all.

## Why pin to a SHA

Referencing `release/v1` is mutable — the tag can be moved to point at a new commit. If the action repository is ever compromised (or a new release introduces a supply-chain risk), a build on our side could silently execute the changed code with permissions to publish a release to PyPI, all without any token in place to protect us. Pinning to an exact commit hash ensures the exact code we reviewed runs, and `dependabot.yml` is configured to open PRs when upstream releases new versions so we can review and update deliberately.

## For review

This depends on @ehpor setting up trusted publishing on PyPI, which relies on me getting access to my account. See pypi/support/issues/11130.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -50,7 +50,7 @@ jobs:
           path: dist/
 
       - name: Publish distribution to PyPI
-        uses: pypa/gh-action-pypi-publish@release/v1
+        uses: pypa/gh-action-pypi-publish@dc37677b2e1c63e2034f94d8a5b11f265b73ba33
 
   attach-to-release:
     name: Attach distribution to GitHub Release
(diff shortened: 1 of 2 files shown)


ehpor:
Added Github as trusted publisher to PyPI account.

syhaffert (approved):
LGTM

PR #352, opened 2026-08-02
Add Array API einsum.

ehpor:
This PR adds a from-scratch, array-API-compatible `einsum` implementation along with cross-backend tests. A large majority of Field operations use einsum in some way, necessitating this new implementation.

## Implementation

`einsum(subscripts, *operands, optimize=True)` is a drop-in replacement for `np.einsum()` (with reduced features) yet working for all Array API backends. The implementation defers to the native implementation for Numpy, Cupy, Jax and Pytorch. As a generic fallback, `opt_einsum` (new dependency) decomposes the contraction into a series of pairwise contractions. Each pairwise step classifies indices into batch/contracted/free groups, reshapes both operands to 3D and delegates to a single batched matmul (or an elementwise multiply when there is no contracted index), then reshapes back. Repeated indices within one operand are collapsed with a diagonal helper that uses the backend's native `diagonal` when available.

This PR also fixes some behavior for the random generators, involving scalars. I also added a test to specifically target that bug.

## Tests

The parametrized case list covers 41 equations (matmul, batched contractions, diagonals, implicit output, scalars, empty dimensions, up to five operands), each run with both optimization modes across the `xp` fixture and compared against `np.einsum` on the same data. Additional tests cover dtype preservation across four dtypes, integer inputs, and scalar results.

## AI

This PR made use of Deepseek V4 Flash and Pro, along with a painful amount of rewriting for implementation simplicity.

### DIFF
--- a/hcipy/_math/random.py
+++ b/hcipy/_math/random.py
@@ -98,7 +98,7 @@ def _torch_gamma(scale=1.0, shape=1.0, size=None, generator=None):
 
 def _normalize_size(size):
     if size is None:
-        return None
+        return ()
     elif not isinstance(size, tuple):
         return (size,)
     return size
@@ -307,7 +307,7 @@ class RandomGeneratorTorch(RandomGenerator):
     def normal(self, mean=0.0, std=1.0, *, size=None):
         size = _normalize_size(size)
 
-        if size is None or len(size) == 0:
+        if len(size) == 0:
             return self._xp.randn(1, generator=self._rng)[0] * std + mean
         else:
             return self._xp.randn(*size, generator=self._rng) * std + mean
@@ -315,7 +315,7 @@ class RandomGeneratorTorch(RandomGenerator):
     def poisson(self, lam=1.0, *, size=None):
         size = _normalize_size(size)
 
-        if size is None or len(size) == 0:
+        if len(size) == 0:
             lam_tensor = self._xp.ones(1) * lam
             return self._xp.poisson(lam_tensor, generator=self._rng)[0]
         else:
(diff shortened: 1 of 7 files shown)


hcipy/_math/random.py
```diff
@@ -257,27 +257,27 @@ def copy(self):
 
     def normal(self, mean=0.0, std=1.0, *, size=None):
         size = _normalize_size(size)
-        return self._rng.normal(mean, std, size)
+        return self._xp.asarray(self._rng.normal(mean, std, size))
```
syhaffert, line 260: Isn't it better to maybe implement these as a decorator? Then it's easier to implement changes later?

syhaffert (commented):
Only got 1 comment on this PR.

hcipy/_math/random.py
ehpor: I'm not sure what you mean. Adding `asarray()` here is to always return an array from this function.

When `size=None` or `size=(1,)`, Numpy returns a scalar. Jax and others return an Array (as they need to support differentiability and so need their own objects rather than Python scalars). To simplify downstream code, I'm forcing output as an Array, even for Numpy, so that downstream code doesn't need to figure out whether it got a Python scalar or Array; it now always receives an Array.

ehpor:
@syhaffert Fixed merge conflicts and replied to comment.

hcipy/_math/random.py
syhaffert: Okay. solved.

syhaffert (approved):
Ready for merging.

PR #353, opened 2026-08-03
Make field operations compatible with Array API.

ehpor:
## Summary

This PR makes HCIPy's field operations work with the new array-API backend. This makes all tests of the core of HCIPy work for both new and old-style fields. All tensor operations (`field_einsum`, `field_dot`, `field_inv`, `field_svd`, `field_transpose`, `field_conjugate_transpose`, `field_determinant`, `field_adjoint`, `field_cross`, and the regularized inverses) now dispatch through the array-API namespace inferred from the field data — including a shared `_normalize_field()` helper for unwrapping new-style field data.

The field-operation tests rewritten to be backend-agnostic, exercising the operations against both `array_api_strict` and numpy backends and asserting equivalence to elementwise reference computations.

### DIFF
--- a/hcipy/field/operations.py
+++ b/hcipy/field/operations.py
@@ -1,6 +1,19 @@
-from .field import Field
+from .field import Field, NewStyleField, is_field
 import numpy as np
 import string
+from .._math.einsum import einsum
+from .._math.backends import array_namespace, is_scalar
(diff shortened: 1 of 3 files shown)


syhaffert (dismissed):

PR #355, opened 2026-08-04
Add tests for performing Fourier transforms on tensor fields.

ehpor:
Add a test for the correctness of Fourier transforms over tensor fields. The Fourier transform should broadcast over the tensor dimensions, as if each component is Fourier transformed independently.

This PR also fixes a bug in `ZoomFastFourierTransform` which performed its transforms around the wrong axes when given a non-scalar field.


syhaffert (approved):

PR #357, opened 2026-08-04
Add keystone basis modes

GeorgeVassilakis:
Hey Emiel,

Following up on the fair warning I gave you at the end of #306: this PR adds an orthonormal mode basis for EAC-2's outer keystone segments, analogous to the Hexike basis in #283. A few design decisions:
- I added `make_keystone_basis()`, along with `keystone()`, `keystone_noll()`, and `keystone_ansi()`, following the same API as the existing Zernike and Hexike functions.
- I exposed `make_wedge_aperture()` through the public aperture API so users can construct the exact mask corresponding to the basis.
- Also fixed a documentation discrepancy I noted in the `make_wedge_aperture()` function which described the diameter parameter as a radius.

With regard to EAC2, it has a circular core, but that should be covered by the standard Zernike modes. The rest are keystone elements that are covered by these new functions.

I tried to keep this as close as possible to the style of the Hexike implementation we integrated previously and avoid unrelated code churn. Happy to tweak the naming or interface if you would prefer anything different. Let me know what you think!

Pinging Susan Redmond (@sfr348) and Maya Amit (@MayaAmit) as they wanted this feature.


ehpor:
I'll have to go through the code still, but just did a quick check visually. The modes look correct.

image

ehpor (approved):
LGTM. No comments.

Great job! 🎉

PR #358, opened 2026-08-05
Native batching for all Fourier transforms.

ehpor:
## Summary

This PR adds native support for tensor fields to all Fourier transforms. Previously, tensor fields (e.g. polarization vectors) were handled by the `multiplex_for_tensor_fields` decorator, which looped over the tensor components and called the scalar transform once per component, then stacked the results. All Fourier transforms now handle the tensor axes internally in a single pass, removing the decorator entirely and avoiding per-component and copying overhead.

This PR is also using `scipy.linalg.get_blas_funcs()` rather than blas directly.

### DIFF
--- a/hcipy/fourier/naive_fourier_transform.py
+++ b/hcipy/fourier/naive_fourier_transform.py
@@ -1,8 +1,10 @@
 import numpy as np
+import math
 
 from .fourier_transform import FourierTransform, ComputationalComplexity, _get_float_and_complex_dtype
 from ..field import Field
 from ..config import Configuration
+from .._math.einsum import einsum
 
 class NaiveFourierTransform(FourierTransform):
     '''The naive Fourier transform (NFT).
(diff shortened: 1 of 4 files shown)


hcipy/fourier/naive_fourier_transform.py
```diff
@@ -93,14 +92,14 @@ def forward(self, field):
             The Fourier transform of the field.
         '''
         if self.precompute_matrices:
-            res = self.matrix_forward.dot(field.ravel())
+            res = np.einsum('ij,...j->...i', self.matrix_forward, field)
```
syhaffert: These need to be replaced with the array-api agnostic einsum, right?

syhaffert (commented):
Looks good to me. Just 1 comment.

hcipy/fourier/naive_fourier_transform.py
ehpor: Once #352 has been merged, yes.

ehpor: Fixed in c527833433497d844a13f62662b922353670458c.

PR #360, opened 2026-08-09
Add linearization of optical systems.

ehpor:
## Summary

Simulation of wavefront sensing and control (e.g. coronagraphic contrast prediction) often requires the sensitivity of the output electric field to (wavefront) error modes. This PR adds functionality to compute these sensitivity matrices and linearized models for you. It adds exact first-order linearization of optical systems with respect to wavefront error modes, plus easier composition of optical elements with the `@` operator.

@ivalaginja @laurentpueyo FYI

## New function and class

- `linearize_optical_system(optical_system, aperture, modes, wavelength=1, mode_type='opd')`: exact first-order linearization of an optical system: returns `E_out = E_0 + G·φ` by propagating each WFE mode once (no finite differences). `mode_type` sets mode semantics: `'opd'`, `'phase'`, `'amplitude'`, or `'field'` for optical path difference, phase-only, amplitude-only and electric field aberrations.
- `LinearizedOpticalSystem`: dataclass holding `static_field` ($$E_0$$), `field_response` ($$G$$), `modes`, `wavelength`, `mode_type`, with `input_grid`/`output_grid` properties and:
  - `__call__(φ)`: evaluate the linearized output field.
  - `intensity_from_covariance(P)`: mean intensity `|E_0|² + diag(G P Gᴴ)` for coefficient covariance `P` (vector or matrix), in `O(n_out·r)` storage.
  - `pastis_matrix(weights)`: weighted PASTIS matrix `Gᴴ W G` (real, symmetric), for contrast prediction via `trace(M P)`.

I also added easier composition for optical elements and systems using the `@` matmul operator. Composition is associative but not commutative. Note: `OpticalSystem` no longer subclasses `OpticalElement`, keeping the two interfaces independent.

Tests were added for all functionality. I used Deepseek V4 Flash and Pro for writing docstrings and tests.

## Todo

- [X] ~Consider deprecation of `get_transformation_matrix_forward()` and `get_transformation_matrix_backward()`. These functions are a weak and memory intensive alternative to the functionality that is introduced in this PR.~ Made #363 to deprecate this if we want to.

### DIFF
--- a/hcipy/optics/__init__.py
+++ b/hcipy/optics/__init__.py
@@ -38,6 +38,8 @@ __all__ = [
     'make_agnostic_forward',
     'make_agnostic_backward',
     'OpticalSystem',
+    'LinearizedOpticalSystem',
+    'linearize_optical_system',
     'PeriodicOpticalElement',
     'jones_to_mueller',
     'JonesMatrixOpticalElement',
@@ -71,6 +73,7 @@ __all__ = [
 ]
 
 from .optical_element import *
+from .linearization import *
 from .wavefront import *
 
 from .aberration import *
--- a/hcipy/optics/linearization.py
+++ b/hcipy/optics/linearization.py
@@ -1,0 +1,205 @@
+from dataclasses import dataclass
+
+import numpy as np
+
+from ..field import Field, FieldBase
+from ..mode_basis import ModeBasis
+from .wavefront import Wavefront
+
+
+@dataclass
+class LinearizedOpticalSystem:
+    '''The first-order (linearized) response of an optical system to wavefront error modes.
+
+    This model describes the output electric field of an optical system to
+    first order in the coefficients `phi` of a set of wavefront error modes:
+
+    `E_out = static_field + field_response * phi`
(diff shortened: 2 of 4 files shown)


syhaffert:
This implicitly assumes that we are interested in the electric field linearization. In principle the implementation looks fine, I just wonder if we want to add an option for an intensity linearization version? But that might be better for a separate PR.

ehpor:
I'm sure there is a need for first-order sensitivities, and they can be derived from the second-order sensitivities (ie. electric field response). But first-order sensitivities are much harder to reason about than these simpler second-order. 

This PR is to support PASTIS matrix calculations and facilitate analyses like Potier et al. 2021 for LUVOIR. We've been using those analyses a lot during the CDS project, so that warranted a more generalized calculation like is done here.

syhaffert (dismissed):
Okay sounds good. Then I think its ready for merging.

ehpor:
@syhaffert Fixed merge conflict. Needs reapproval.

PR #362, opened 2026-08-12
Strip linebreaks before installing optional dependencies

ehpor:
Fixes optional-dependency install on Windows where the file is checked out with CRLF line endings, causing conda to reject specs like `mkl_fft\r`.

- Pipe `optional_dependencies.txt` through `tr -d '\r'` in both workflows (`testing.yml`, `array_api.yml`)
- Quote `$requirement` when passing to conda


raphaelpclt (approved):
Ok

PR #363, opened 2026-08-12
Deprecate Fourier and OpticalElement transformation matrix methods

ehpor:
Deprecate `get_transformation_matrix_forward()` and `get_transformation_matrix_backward()` on `OpticalElement` and its child classes and on the `FourierTransform` class. Corrolary is the deprecation of `generate_app_por()`. I do not think any people are using this functionality.

These transformation matrix methods are extremely memory-intensive, too much for practical use, and are only viable for toy models or for small grids / big machines.

See #360 for functionality that replaces it, but in a much better way.


syhaffert (dismissed):
Agree that these methods are almost never used. I also don't think the generate_app function was ever used. Approved for merging.

PR #371, opened 2026-08-14
Fix spider sign

### LINKED ISSUE #238: Sign error in reference point in make_spider_infinite
While working on the new Magellan aperture #237 , I found that the make_spider_infinite has a sign error in the starting coordinate. I will fix this in a separate pull request because this requires me to go through all the other apertures that use the make_spider_infinite and switch the sign.

syhaffert:
PR to fix the bug mentioned in #238 . This will allow us to close that issue. This PR does create an interface change so people that have implemented their own apertures in local codes will need to switch the sign after upgrading to the new version.

I ran all the aperture tests and they all still pass.

Fixes #238.

### DIFF
--- a/hcipy/aperture/generic.py
+++ b/hcipy/aperture/generic.py
@@ -376,8 +376,8 @@ def make_spider_infinite(p, angle, spider_width):
         else:
             x, y = g.coords
 
-        x = x + p[0]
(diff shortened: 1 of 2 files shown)


ehpor (approved):
LGTM.

PR #372, opened 2026-08-15
Lyot fpm fix

### LINKED ISSUE #177: Bug in LyotCoronagraph if used with focal plane OpticalElement

syhaffert:
This PR fixes the bug from #177 . This PR was developed together with MAI-Code-1.1-Flash.

This PR adds a new function `get_optical_element_input_grid` that tries to determine the input grid from the optical element. It works on all apodizers and polarization optics. This will remove the need to explicitly passing the coronagraph grid in the LyotCoronagraph. I kept the explicit parameter `focal_plane_mask_grid` to keep backwards compatibility.

Several test functions have been added to test the interface for various optical elements.

Let me know if this is an agreeable solution, I really get messed up by this particular bug very often.

Fixes #177.

### DIFF
--- a/hcipy/optics/__init__.py
+++ b/hcipy/optics/__init__.py
@@ -37,6 +37,7 @@ __all__ = [
     'AgnosticOpticalElement',
     'make_agnostic_forward',
     'make_agnostic_backward',
+    'get_optical_element_input_grid',
     'OpticalSystem',
     'PeriodicOpticalElement',
     'jones_to_mueller',
--- a/hcipy/optics/__init__.py
+++ b/hcipy/optics/__init__.py
@@ -37,7 +37,6 @@ __all__ = [
     'AgnosticOpticalElement',
     'make_agnostic_forward',
     'make_agnostic_backward',
-    'get_optical_element_input_grid',
     'OpticalSystem',
     'PeriodicOpticalElement',
     'jones_to_mueller',
--- a/hcipy/optics/optical_element.py
+++ b/hcipy/optics/optical_element.py
@@ -818,7 +818,7 @@ def _get_optical_element_input_grid(optical_element):
             return grid
     except (AttributeError, TypeError, ValueError):
         pass
-    
+
     for attribute_name in [
         'grid',
         'apodization',
--- a/hcipy/optics/optical_element.py
+++ b/hcipy/optics/optical_element.py
@@ -789,12 +789,13 @@ def make_agnostic_backward(backward):
         return backward(self, instance_data, wavefront, *args, **kwargs)
     return res
 
-def get_optical_element_input_grid(optical_element):
+def _get_optical_element_input_grid(optical_element):
     '''Get the input grid of an optical element.
 
-    This function tries to get the input grid of an optical element by
-    checking its attributes and methods. If the input grid cannot be
-    determined, a ValueError is raised.
+    This function checks a small, explicit list of the grid-bearing properties that
+    are used throughout this module. The order is intentional: direct input-grid
+    information is preferred, followed by the most common field-like properties that
+    carry their own grid metadata.
 
     Parameters
     ----------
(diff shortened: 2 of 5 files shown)


hcipy/optics/optical_element.py
```diff
@@ -789,6 +789,56 @@ def res(self, wavefront, *args, **kwargs):
...
+
+        if isinstance(value, dict):
+            stack.extend(value.values())
+        elif isinstance(value, (list, tuple, set)):
+            stack.extend(value)
+        else:
+            try:
+                stack.extend(vars(value).values())
+            except TypeError:
+                pass
```
ehpor: This seems to recursively loop through all __dict__ elements of the optical element (not properties, I might add), in order to find one that has a `grid` attribute (which may or may not be the input grid). That is really necessary?! This seems errorprone and overly complicated.

hcipy/optics/__init__.py
```diff
@@ -37,6 +37,7 @@
     'AgnosticOpticalElement',
     'make_agnostic_forward',
     'make_agnostic_backward',
+    'get_optical_element_input_grid',
```
ehpor: I strongly disagree with exporting this function. This feels like an internal helper function, waiting for a better solution that takes more time to generalize.

hcipy/optics/optical_element.py
syhaffert: Yeah, I agree. I worked on simplifying the function by making it very explicit in what it checks. The only optical element that it will fail on will be the MultiplexedComplexSurface, which I think no one uses anyway.

hcipy/optics/__init__.py
syhaffert: Yes, I originally added it to make sure I could import it directly from ..optics. I have now removed it.

hcipy/optics/optical_element.py
```diff
@@ -789,6 +789,62 @@ def res(self, wavefront, *args, **kwargs):
...
+
+        if value is None or callable(value):
+            continue
+
+        if hasattr(value, 'grid'):
+            grid = value.grid
+            if grid is not None and not callable(grid):
+                return grid
+
+    try:
```
ehpor: This check should be checked first, no? Like, if the class exposes this and gives a coherent answer, then that is truth. Otherwise, we'll walk the list of specific properties.

ehpor (changes requested):
One comment, otherwise okay.

hcipy/optics/optical_element.py
syhaffert: Agreed. I switched them.

ehpor (approved):

PR #373, opened 2026-08-15
Separable filters for the FFT shift operations

ehpor:
## Summary

Replaces the numexpr-based shift filters in the Fourier transforms with a new `SeparableFilter` class (`hcipy/_math/separable_filter.py`), built on per-axis 1-D factors.

## Changes

- **`SeparableFilter` dataclass**: one 1-D factor per axis; `threshold` decides at apply time between the full N-D filter array (single numpy multiply) and a fused numba kernel (single pass over the factors). `apply()` dispatches to `apply_numpy(out=)` for numpy arrays; `inverse=True` divides by the filter. Generated kernels broadcast over leading tensor (batch) axes.
- **Factories**: `make_phase_ramp(slopes, grid)` and `make_separable_filter(factors)`.
- **`FastFourierTransform`**: the shift filters (native-shift ramp, fftshift emulation, piston removal, normalization weights) are built once per side via `_make_shift_filter` and applied with `apply_numpy(..., out=)` inside forward/backward; `_numexpr_grid_shift` and the `numexpr` dependency for this path are removed. Compatible with the tensor-field batching of `feature/fourier_batching`.
- **`FourierShift`**: builds its filter from per-axis `ifftshift`ed factors on the full FFT grid; this also fixes a latent bug where shifts with a zero first element (e.g. `[0, sy]`) silently produced a ones filter.
- **Tests**: roundtrip, out=/in-place, inverse, batched-apply and kernel tests; `numba` added as a dependency.

## Performance

At parity for small grids and large grids without emulated fftshifts; ~1.8–2.4× faster forward/backward for large grids with `emulate_fftshifts=True` (fused numba kernel), and faster construction.


ehpor:
## FFT benchmark: SeparableFilter shift filters vs the previous implementation

This benchmark compares `FastFourierTransform.forward` and `backward` between the previous implementation (flat shift filters built with numexpr via `_numexpr_grid_shift`) and the new implementation, which builds the shift filters as `SeparableFilter` objects (per-axis factors, applied by a fused numba kernel above the 65536-point threshold and by a full-array multiply below it).

### Benchmark parameters

- **Hardware**: Apple Silicon (macOS), CPU only.
- **Software**: numpy 2.4.6, numba 0.65.1 (12 threads), scipy for the FFT pipeline checks.
- **Input**: complex128 fields on uniform grids of N x N pixels, N in {256, 512, 1024, 2048}.
- **`emu`**: the `emulate_fftshifts` option of the Fourier transform, off/on.
- **Timing**: `forward` and `backward` are timed *standalone*, as the minimum of 5 repeats inside a 0.3 s window, with garbage collection in between. Both implementations run in the same process.
- **Correctness**: the outputs of the two implementations were verified to agree to ~1e-14.

### Results

| N | emu | fwd old (ms) | fwd new (ms) | fwd speedup | bwd old (ms) | bwd new (ms) | bwd speedup |
|---|---|---|---|---|---|---|---|
| 256 | off | 0.389 | 0.383 | 1.02x | 0.440 | 0.439 | 1.00x |
| 256 | on | 0.334 | 0.305 | 1.10x | 0.420 | 0.399 | 1.05x |
| 512 | off | 1.805 | 1.901 | 0.95x | 2.127 | 1.855 | 1.15x |
| 512 | on | 1.617 | 0.889 | 1.82x | 1.886 | 0.910 | 2.07x |
| 1024 | off | 4.698 | 4.322 | 1.09x | 5.485 | 4.242 | 1.29x |
| 1024 | on | 5.485 | 2.294 | 2.39x | 5.762 | 2.360 | 2.44x |
| 2048 | off | 24.735 | 23.138 | 1.07x | 27.303 | 22.185 | 1.23x |
| 2048 | on | 19.423 | 11.419 | 1.70x | 23.279 | 11.506 | 2.02x |

### Discussion

- Without emulated fftshifts, large grids are at parity to moderately faster (forward 0.95-1.09x; backward 1.15-1.29x for N >= 512). This is due to the Numba kernel not needing to allocate and calculate the inverse of the shift arrays.
- With emulated fftshifts, the fused numba kernel replaces the flat numpy multiply of the shift filters, giving the largest gains: 1.7-2.4x for forward and ~2.0-2.4x for backward at N >= 512.
- The only configuration slower than parity is the forward at 512 with `emulate_fftshifts=False` (0.95x, i.e. about 5% slower), where the kernel launch overhead outweighs the kernel speed at this size; at 1024 and above the kernel is faster than the numpy multiply.
- Filter construction (not shown) is also faster, since the filters store 1-D factors rather than full shift arrays for large grids.

ehpor:
## Design space for the `SeparableFilter` class

I benchmarked the apply step of a 2D phase ramp (complex128, N from 64 to 4096) across the full matrix of options: 3 output modes (in-place / preallocated `out=` / fresh allocation) x 5 engines (numpy, numexpr with 1 or 12 threads, numba serial, and numba parallel) x 2 forms (`2D`: single multiply by a precomputed full N-D phase array; `sep`: per-axis 1-D factors). The options differ in two dimensions:

- **Memory vs passes**: the `2D` form precomputes the full N-D phase array once, so each application is a single multiply -- but it costs N^2*16 bytes at setup; the `sep` form keeps only 1-D factors but needs either several broadcasting multiplies (numpy) or a single fused pass (numexpr/numba).
- **Parallelization**: numexpr 1T and numba serial are consistently the slowest; numexpr 12T and numba parallel are the fastest separable engines, closely matched; numpy (single-threaded) is best for the `2D` form, where a single well-vectorized pass beats the threading overhead.

The takeaway was regime-dependent: below ~65536 points the `2D` full-array multiply wins (the kernel overhead and the extra passes don't pay off), while above it the fused parallel kernels win by 3-4x over numpy's separable two-step and comfortably beat the `2D` form once the full array becomes memory-bound.

`SeparableFilter` reproduces the best of both regimes automatically: it stores per-axis factors, and the `threshold` switches between exactly these two paths -- a single full-array multiply below 65536 points, and a fused numba-parallel kernel (the fastest separable engine, without numexpr's per-call thread-pool setup) above it. `apply_numpy(out=)` closes the allocation gap shown by the fresh-alloc vs `out=` rows, and the FFT builds its shift filters once at init, so per-application cost matches the benchmarked `sep` row. The kernel is generated at runtime for any number of dimensions, with a leading batch axis so tensor fields are handled in the same fused pass.

syhaffert (dismissed):
Looks good to me. Ready for merging after merge conflicts have been solved.

ehpor:
@syhaffert Fixed merge conflicts.

syhaffert (dismissed):

PR #376, opened 2026-08-19
Add `FraunhoferPropagator.closest_fft_wavelength`.

ehpor:
## Summary

Adds `FraunhoferPropagator.closest_fft_wavelength(target_wavelength)` and tests for it.

## Motivation

For a monochromatic Fraunhofer propagation the uv-grid is the output grid scaled by `2π / (f · λ)`. This grid is a *native FFT grid* of the input grid only when, for every axis, the number of zero-padding pixels

```
zp = f · λ / (input_grid.delta[i] · output_grid.delta[i])
```

is an integer of at least `max(input_grid.dims, output_grid.dims)`. In that case `make_fourier_transform` _can_, but doesn't have to, select the fast `FastFourierTransform`; otherwise it falls back to a `MatrixFourierTransform`.

Using FFTs _can_ be faster than MFTs and using the correct wavelength enables this potential performance improvement, provided that the user is amenable to changing their wavelengths slightly.

I used Deepseek V4 Flash to generate a draft of the code which was then rewritten.

### DIFF
--- a/hcipy/propagation/fraunhofer.py
+++ b/hcipy/propagation/fraunhofer.py
@@ -1,8 +1,8 @@
-import numpy as np
-
 from ..optics import Wavefront, AgnosticOpticalElement, make_agnostic_forward, make_agnostic_backward
 from ..field import Field
 from ..fourier import make_fourier_transform
+
+import math
 
 class FraunhoferPropagator(AgnosticOpticalElement):
     '''A monochromatic perfect lens propagator.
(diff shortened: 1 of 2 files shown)


syhaffert (approved):
I approve this as long as this stays a separate function that users can call if they understand what they are doing.

ehpor:
Nope, this is supposed to be a power-user only function. Requested wavelengths should always stay exact, but I use this a lot to give ~2-3x higher performance for 0.5nm in wavelength shift (when I don't care about the exact wavelengths).

Issue #109, opened 2021-07-18
Planar to spherical propagation


syhaffert:
For several end-to-end instrument simulations it would be helpful if there is a spherical to planar propagator based on Fresnel diffraction. I think the Zemax documentation has a pretty good description of the different transforms and regions. I wanted to open an issue to discuss the implementation. 

I was thinking of adding a function like:
`propagator = choose_propagator(wavefront_in, distance)`

I have assumed that the wavefront_in has an internal Gaussian beam that keeps track of the divergence during all propagations. Then based on the divergence of the Gaussian beam it selects Fresnel/ASP or the SphericalFresnelPropagator.

For the propagator itself, we can just add SphericalFresnelPropagator to differentiate it from the Planar version. And I think it is best to _**not**_ rename the old propagator to keep backwards compatibility.

Any thoughts?

ehpor:
My original plan for this was to have a `Propagator` child class that did this, without any function like the one you propose. That class would decide on the propagator to use based on its input wavefront (including the pilot Gaussian beam). It would internally use the `FresnelPropagator` and `FraunhoferPropagator` to propagate from plate to spherical and spherical to plane (via the waist of the Gaussian beam). Why do you need your proposed function? It avoids the need for internal storage of the propagators, but it complicates user code.

FYI, something related: I'd like to revert to the old way of doing the agnostic optical elements. I realize now that having the optical element be agnostic to input grid was a mistake and overcomplicates things. I'm currently using an intermediate solution myself, where the input grid and future Gaussian beam are fixed after (lazy) initialization, but optical elements should still be able to handle input wavefronts with different wavelengths. That has worked out so far, and I aim to push that after SPIE (and my holiday). It simplifies internal hcipy code, especially for the planned multi-backend support, and is almost equivalent for user code.

I wholeheartedly agree with never renaming classes. Having aliases however is always an option.

syhaffert:
A child class works for me too. The planar to spherical (and vice-versa propagation) is a bit different from the Fraunhofer propagation, right? It's `factor * quadratic_phase(output_grid) * Fourier Transform{ quadratic_phase(input_grid) }`.
The quadratic phases have to be added and the spatial frequency sampling is different? Or is the sampling equal if the propagation distance is chosen instead of the focal length?

Yeah I think the current agnostic implementation is indeed a little bit over engineered.

philhinzatucsc:
[...]

PR #377, opened 2026-08-19
Add ray transfer (ABCD) matrices and Gaussian beam propagation

ehpor:
## Add ray transfer (ABCD) matrices and Gaussian beam propagation

Adds a backend-agnostic 2x2 ray-transfer (ABCD) matrix module plus functional Gaussian beam propagation, usable with any Array API backend. Also refactors the `GaussianBeam` class to use the complex beam parameter as main variable.

### Features
- `RayTransferMatrix` wraps a fixed 2x2 array or a wavelength callable (dispersive matrices evaluated at the requested wavelength); supports `@` composition and derived scalars (`effective_focal_length`, `paraxial_power`, `distance`, `front/back_focal_length`, ...) plus `is_surface`/`is_free_space` predicates.
- Constructors: `make_abcd_free_space`, `make_abcd_thin_lens`, `make_abcd_mirror`, `make_abcd_fraunhofer`, `make_abcd_magnifier`, `make_abcd_refractive_interface`.
- `RayTransferMatrix.forward(beam)` / `backward(beam)` transform a `GaussianBeam` via the ABCD law `q' = (Aq+B)/(Cq+D)`, returning a new beam; `backward` applies the numerical inverse.
- `GaussianBeam.propagate(matrix)` accepts a raw matrix or callable (evaluated at the beam's wavelength) and returns a new propagated beam.

### Design notes
Physical-angle convention: (`u = dy/dz`). Under this convention determinants are not restricted to 1. This allows the Gaussian beam to keep track of its refractive index without explicitly telling it what the new one is.

### Testing
All new tests run against `array_api_strict` only, just like the other Array API compatible tests.

### DIFF
--- a/hcipy/optics/abcd.py
+++ b/hcipy/optics/abcd.py
@@ -102,6 +102,50 @@ class RayTransferMatrix:
 
         return (abs(a - 1.0) < tol) and (abs(c) < tol) and (abs(d - 1.0) < tol)
 
+    def forward(self, beam):
+        '''Transform a Gaussian beam through this matrix, returning a new beam.
+
+        The complex beam parameter of the beam is transformed according to the
+        ABCD law
+
+            q' = (A q + B) / (C q + D).
+
+        A new GaussianBeam is returned; the input beam is not modified. A
+        dispersive matrix is evaluated at the wavelength of the beam.
+
+        Parameters
+        ----------
+        beam : GaussianBeam
+            The Gaussian beam to transform.
+
+        Returns
+        -------
+        GaussianBeam
+            The transformed Gaussian beam.
+        '''
+        return beam.propagate(self(beam.wavelength))
+
+    def backward(self, beam):
+        '''Inverse-transform a Gaussian beam through this matrix.
+
+        The numerical inverse of the matrix is applied, so this works for any
+        invertible matrix.
+
+        Parameters
+        ----------
+        beam : GaussianBeam
+            The Gaussian beam to transform.
+
+        Returns
+        -------
+        GaussianBeam
+            The inverse-transformed Gaussian beam.
+        '''
+        M = self(beam.wavelength)
+        xp = infer_xp(M)
+
+        return beam.propagate(xp.linalg.inv(M))
+
     def __matmul__(self, other):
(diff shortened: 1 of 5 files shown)


ehpor:
After reviewing on my phone, I think we also need to have a refractive index and need to track it when propagating the gaussian beam.

ehpor:
Took the time to refactor `GaussianBeam`. It is now using `q` as the main parameter, is immutable (easier to reason about), includes and tracks the refractive index, and now supports the Array API. It still returns Python scalars if it's only given Python scalars as input.

hcipy/optics/abcd.py
```diff
@@ -0,0 +1,334 @@
...
+    'make_abcd_free_space',
+    'make_abcd_thin_lens',
+    'make_abcd_mirror',
+    'make_abcd_fraunhofer',
+    'make_abcd_magnifier',
+    'make_abcd_refractive_interface',
+]
+
+class RayTransferMatrix:
+    '''A 2x2 ray-transfer (ABCD) matrix, optionally dispersive.
```
syhaffert, line 14: Can you add which definition you are following? This should be either [x, theta] or [theta, x].

hcipy/optics/abcd.py
```diff
@@ -0,0 +1,334 @@
...
+
+    Wraps either a fixed 2x2 array or a function of the wavelength (in
```
syhaffert, line 16: Do we only assume circular symmetric systems because of the 2x2 assumption? Instead of the 4x4 to support (x, y, theta_x, theta_y) ?

syhaffert (commented):
Two comments about the implementation.

hcipy/optics/abcd.py
ehpor: I did not know that [theta, x] is used by anyone. Do you want that convention just here or for all ABCD creation functions?

hcipy/optics/abcd.py
ehpor: I don't have a need for it. This is mostly to prevent aliasing for a future implementation of #109, which is gonna use circularly symmetric matrices. You have a strict need for non-circularly-symmetric systems?

PR #378, opened 2026-08-20
Rewrite the configuration system with pydantic-settings

ehpor:
### Summary

Replaces the hand-rolled YAML-based configuration with a `pydantic-settings` model. The configuration is now a flat, validated singleton that can be set in code, in the project/user configuration files, or in `HCIPY_*` environment variables — and reacted to via a new `on_config_change` decorator.

### Motivation

The old configuration was a loose dictionary-like tree loaded from YAML, with no validation and no way to know when values changed. Typos and type errors silently slipped through, and settings could only be set in configuration files, not per-session or per-run.

### What changed

**New API**
- `Configuration()` is a flat singleton model with fields such as `fft_method`, `fft_emulate_fftshifts`, `fft_runtime_coeffs`, `mft_precompute_matrices`, `cmap_psf`, `ffmpeg_path`, and `use_array_api`.
- Every assignment is validated (correct type, known field); unknown fields raise `ValidationError` instead of failing silently.
- Precedence: assignments in code > `HCIPY_*` environment variables > `./hcipy_config.yaml` > `~/.hcipy/hcipy_config.yaml` > built-in defaults.
- `Configuration().reset()` re-reads all sources; `reset(enable_user_overrides=False)` runs with defaults only (ignores env vars and config files).
- New `on_config_change` decorator registers callbacks that fire on every assignment and reload (idempotent by design).
- The execution-time prediction coefficients are now `(a, b, c)` tuples instead of `{a, b, c}` dicts.

**Consumers updated** across `_math`, `field`, `fourier`, and `plotting` to the flat field names. The old `hcipy/config/default_config.yaml` is removed; defaults live in the model itself.

**Docs**
- New intermediate-level tutorial **Configuration** (notebook, compiled by the existing doc pipeline) covering what the config is, when to use it, and how to set values in code, in config files, and in environment variables.
- `doc/performance.rst` updated to the new config format (`use_array_api`, flat `*_runtime_coeffs` keys).

### Notes / breaking changes
- `use_array_api` (renamed from `use_new_style_fields`) is still read at import time only. While the functionality is in place to fix that, this PR does not do that since it comes with advantages/disadvantages particular to that issue and irrelevant to the config system itself. Deferring to a separate PR.
- Existing user configuration files (e.g. `~/.hcipy/hcipy_config.yaml`) written in the old nested format with section headers (`fourier:`, `plotting:`, `core:`, …) will now raise a `ValidationError` and must be migrated to the flat key format (`use_array_api: true`, `fft_method: [...]`, …). This is done using the new `hcipy_migrate_config` CLI command.


ehpor:
Fixed merge conflicts.

raphaelpclt (approved):
Ok

PR #379, opened 2026-08-24
Replace wildcard import in tutorials with explicit hcipy namespace.

ehpor:
## Replace wildcard import in tutorials with explicit `hcipy` namespace

All tutorial notebooks previously used `from hcipy import *`. Wildcard imports make it unclear where a name comes from and can silently shadow local variables. This PR replaces that with `import hcipy` and prefixes every call to an hcipy function or class with the `hcipy.` namespace, e.g.

```python
# before
pupil_grid = make_pupil_grid(256)
wavefront = Wavefront(aperture)

# after
pupil_grid = hcipy.make_pupil_grid(256)
wavefront = hcipy.Wavefront(aperture)
```


syhaffert (approved):
Looks good to me. Ready for merging. I think this is an excellent idea. I have also started to not do a wildcard import anymore for exactly the same reason. I think it's good that we try to drill this by including it like this in the tutorials.

PR #387, opened 2026-09-09
Broaden test tolerance and make math tests determinstic.

ehpor:
The tight tolerances led to spurious test failures. Tests also now use a fixed seed for the random generator, making them deterministic, making it easier in the future to debug.


raphaelpclt (approved):
Ok

